PR #26, opened 2021-02-05
Tweak and expand navigation interception

### LINKED ISSUE #3: Clarify "same-document" and "cross-document" terminology in the explainer
Moved from https://github.com/slightlyoff/history_api/issues/31 by @matt-buland-sfdc and summarized by me.

The explainer discusses same-document and cross-document navigations. While this concept is clear to spec people and implementers, it is not commonly used by web developers, and can cause confusion. (Additionally, I've previously gotten feedback that "cross-document" sounds kind of security-related, like "cross-origin", so maybe "different-document" would be a better name.)

The meaning of these terms is that any navigation that would normally result in a new `Document` object is what I call "cross-document". I believe the only "same-document" navigations are: fragment navigations, `history.pushState()`/`appHistory.pushNewEntry()`, and `history.replaceState()`/`appHistory.updateCurrentEntry()`.

This should be summarized the first time the term is used. Additionally, this can help with the confusion around case (5) mentioned in https://github.com/slightlyoff/history_api/issues/19.

domenic:
The normative change here is to resolve "TODO: should the event even fire at all, for these latter two cases?" in the negative: i.e., the navigate event will not fire at all for non-interceptable cases. Additionally, this makes same-document fragment navigations, even initiated via browser UI or window.open(), interceptable.

Beyond that, this commit:

* Expands the discussion of navigation interception to be clearer about what "user-initiated" means
* Adds an appendix detailing all the navigations on the platform, the distinction between cross-document and same-document (closes #3), and a summary table listing how this spec works for each type.
* Discusses the connection between the navigations we're considering, and the HTML Standard algorithms.

/cc @matt-buland-sfdc and @natechapin; I'll probably let this sit for a day or two before merging in case you have any comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ This new `window.appHistory` API layers on top of the existing API and specifica
 - [Security and privacy considerations](#security-and-privacy-considerations)
 - [Stakeholder feedback](#stakeholder-feedback)
 - [Acknowledgments](#acknowledgments)
+- [Appendix: types of navigations](#appendix-types-of-navigations)
 - [Appendix: full API surface, in Web IDL format](#appendix-full-api-surface-in-web-idl-format)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -209,11 +210,11 @@ All of these methods return promises, because navigations can be intercepted and
 
 ### Navigation monitoring and interception
 
-The most interesting event on `window.appHistory` is the one which allows monitoring and interception of navigations: the `navigate` event. It fires on any navigation, either user-initiated or application-initiated, which would update the value of `appHistory.currentEntry`. This includes cross-origin navigations (which will take us out of the current app history list), but it does _not_ include navigations which target other windows, e.g. right click/open in new tab. **We expect this to be the main event used by application- or framework-level routers.**
+The most interesting event on `window.appHistory` is the one which allows monitoring and interception of navigations: the `navigate` event. It fires on almost any navigation, either user-initiated or application-initiated, which would update the value of `appHistory.currentEntry`. This includes cross-origin navigations (which will take us out of the current app history list); see [below](#example-cross-origin-affiliate-links) for an example of how this is useful. **We expect this to be the main event used by application- or framework-level routers.**
 
 The event object has several useful properties:
 
-- `userInitiated`: a boolean indicating whether the navigation is user-initiated (i.e., a click on an `<a>`, or a form submission) or application-initiated (e.g. `location.href = ...`, `appHistory.navigateTo(...)`, etc.).
+- `userInitiated`: a boolean indicating whether the navigation is user-initiated (i.e., a click on an `<a>`, or a form submission) or application-initiated (e.g. `location.href = ...`, `appHistory.pushNewEntry(...)`, etc.). Note that this will _not_ be `true` when you use mechanisms such as `button.onclick = () => appHistory.pushNewEntry(...)`; the user interaction needs to be with a real link or form. See the table in the [appendix](#appendix-types-of-navigations) for more details.
 
 - `destinationEntry`: an `AppHistoryEntry` containing the information about the destination of the navigation. Note that this entry might or might not yet be in `window.appHistory.entries`; if it is not, then its `state` will be `null`.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -803,10 +809,65 @@ Thanks also to
...
 [@housseindjirdeh](https://github.com/housseindjirdeh),
-[@jakearchibald](https://github.com/jakearchibald), and
+[@jakearchibald](https://github.com/jakearchibald),
+[@matt-buland-sfdc](https://github.com/matt-buland-sfdc), and
 [@slightlyoff](https://github.com/slightlyoff)
 for their help in exploring this space and providing feedback.
 
+## Appendix: types of navigations
+
+The web platform has many ways of initiating a navigation. For the purposes of this API, and in particular the [`navigate` event](#navigation-monitoring-and-interception), the following is intended to be a comprehensive list:
```
matt-buland-sfdc, line 819: Could you clarify "this API" so that we know what "this" refers to? For example, I think this would read better: Below is a comprehensive list of ways that a navigation can be initiated, many of which are hooked into the `appHistory` API and the `navigate` event in particular.

README.md
```diff
@@ -803,10 +809,65 @@ Thanks also to
...
+
+- Users triggering navigations via browser UI, including (but not necessarily limited to):
```
matt-buland-sfdc, line 821: Below are mechanisms, rather than actions. A user uses one of these to initiate a navigation. To maintain the current description, I'd expect the list to contain items like "change the text in the URL bar" (to initiate a navigation).

Or, describe these as "mechanisms that can be used to trigger a navigation" (no longer using active voice)

README.md
```diff
@@ -803,10 +809,65 @@ Thanks also to
...
+- `history.pushState()` and `history.replaceState()`
+- `appHistory.back()`, `appHistory.forward()`, `appHistory.navigateTo()`
+- `appHistory.pushNewEntry()` and `appHistory.updateCurrentEntry()`
+- [`document.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/open)
+
+**Cross-document navigations** are navigations where, after the navigation completes, you end up in a different `Document` object than the one you are curently on. Notably, these unload the old document, and stop running any JavaScript code from there.
+
+**Same-document** navigations are ones where, after the navigation completes, you stay on the same `Document`, with the same JavaScript environment.
+
+Most navigations are cross-document navigations. Same-document navigations can happen due to:
```
matt-buland-sfdc, line 842: What can be confusing for many users is that these mechanisms _assume_ they are cross-document, even if they wind up fetching the same document at runtime. Using the wrong mechanism leads to lower page performance and excessive refreshes because they are not integrated into the SPA. I think this section does a good job of describing the expectations, and the real goal for SPAs is to get as many clicks as possible into that beloved 3rd column: `navigate` event.

README.md
```diff
@@ -803,10 +809,65 @@ Thanks also to
...
+- `appHistory.pushNewEntry()` and `appHistory.updateCurrentEntry()`
+- `document.open()`
+- [Intercepting a cross-document navigation](#navigation-monitoring-and-interception) using the `appHistory` object's `navigate` event, and calling `event.respondWith()`
+
+Here's a summary table:
+
+|Trigger|Cross- vs. same-document|Fires `navigate`?|`event.userInitiated`|
+|-------|------------------------|-----------------|---------------------|
+|Browser UI (fragment change only)|Always same|Yes|Yes|
+|Browser UI (other)|Always cross|No|—|
```
matt-buland-sfdc, line 855: Does this include browser back? For entries that were previously pushed, browser back should be equivalent to `history.back()`. So long as the SPA is pushing correctly, back & forward inside the app's history entries should fire `navigate`. (until the user goes back/forward out of the app's history space).

domenic: That doesn't work though, because if you allow intercepting backs, even same-app-history ones, then the page can just trap the user forever on their origin, never letting them actually get back to where they were.

If you want to _see_ backs, you can use `currententrychange`. But since we cannot allow them to be interceptable for anti-abuse reasons, we don't fire `navigate` for them.

README.md
domenic: The below are specific pieces of browser UI, so I think the phrasing is appropriate here? i.e. "... browser UI, including (but not necessarily limited to): the URL bar, ..."

README.md
matt-buland-sfdc: Is it worth adding another column for `currententrychange`? I don't want this nuance to be lost

README.md
matt-buland-sfdc: Yeah, I see. Maybe just change "triggering" to "can trigger" then?

README.md
domenic: I should try to incorporate it somehow (beyond what's already mentioned in the main section: "These restrictions are designed so that canceling..."). But a dedicated column isn't great, because its values are basically the same as the same- vs. cross-document column. That is, cross-document navigations don't fire `currententrychange` (since by the time the entry has changed, the document is gone, and no event listeners fire on the old document). And same-document navigations always do.

matt-buland-sfdc: Yeah, the back/capture part is clear in the other section, I just rabbit-holed on what this table says on its own.

I think https://github.com/WICG/app-history#current-entry-change-monitoring is also descriptive enough to understand when I'd expect `currententrychange`, even if it's not listed in table format. But just to make sure I do understand... If I do a cross-document navigation from A.foo/ to B.foo/ via a URL-bar edit, then the A document does not hear `navigate` nor `currententrychange`, but the B document does hear a `currententrychange`. Is that correct?

domenic: This discussion made me realize that we can't intercept the back button even for fragment changes, so I pushed some more changes in that direction...

matt-buland-sfdc: yes, that's a good clarification. I had assumed you were only describing editing the fragment in the URL-bar (a forward fragment change)

domenic: > But just to make sure I do understand... If I do a cross-document navigation from A.foo/ to B.foo/ via a URL-bar edit, then the A document does not hear navigate nor currententrychange,

Correct

> but the B document does hear a currententrychange. Is that correct?

Well... yes, as currently written, but I'm not sure that's a good idea. See the TODO in https://github.com/WICG/app-history/pull/27. In particular, this would be the only time `currententrychange` fires for a cross-document navigation. And it'd be basically redundant with the `load` event. It's a weird special case, and I suspect firing it might hurt people more than it helps. Probably this should get its own dedicated issue...

matt-buland-sfdc: Started an issue thread: #30 .

domenic:
@natechapin I'm now wondering about the case of `history.back()` from within an iframe. In particular, consider the following joint session history:

```
A. https://example.com/start
B. https://example.com/outer
   ┗ https://example.com/inner
```

Let's say the code in `https://example.com/inner` calls `history.back()`. This should definitely be interceptable by `https://example.com/inner`. But should it be interceptable by `https://example.com/outer`?

Preventing such navigation is not important for anti-abuse reasons (since it's not the user initiating the back). However, I don't believe it's something you could do today. So adding such interception would be expanding capabilities. Do you have any thoughts? In general (see #32), web developers are pretty sad about any restrictions we place on interception.

A natural followup is that `window.open(url, "nameOfOtherWindow")` is very similar: not something you can intervene on today, but, intervening on it is not important for anti-abuse. So we might want to decide the same way for both.

natechapin:
> Let's say the code in `https://example.com/inner` calls `history.back()`. This should definitely be interceptable by `https://example.com/inner`. But should it be interceptable by `https://example.com/outer`?
> 
> Preventing such navigation is not important for anti-abuse reasons (since it's not the user initiating the back). However, I don't believe it's something you could do today. So adding such interception would be expanding capabilities. Do you have any thoughts? In general (see #32), web developers are pretty sad about any restrictions we place on interception.

chromium doesn't know which frame(s) will be navigated by `history.back()` synchronously. My guess is it would be possible to have the parent intercept the navigation, but it'll definitely be additional plumbing and an additional point from which we'll need to fire the navigate event.

> 
> A natural followup is that `window.open(url, "nameOfOtherWindow")` is very similar: not something you can intervene on today, but, intervening on it is does not allow abuse. So we might want to decide the same way for both.

Agreed that this is similar.

domenic:
After some offline discussion with @natechapin, I've updated this PR to allow navigation interception for cases where the navigation is initiated by code in another frame. This goes beyond what is possible to intercept today, but does not have abuse implications, so in theory it should be OK. We may need to scale back this ambition if we find it too hard to implement.

PR #27, opened 2021-02-05
Rename and update semantics for navigation duration measurement

### LINKED ISSUE #8: Consider replacing navigationStartTimeStamp with navigationStart
This would match the performance timing record naming. ex. https://developer.mozilla.org/en-US/docs/Web/API/Navigation_timing_API

### LINKED ISSUE #12: Need timestamp from the initiation of the navigation, not the navigation event
For reporting user interaction metrics we need the start of the user action, not the start of the navigation.

For example if a user clicks the Next button 10 times rapidly (per the photo gallery example), the `navigate` event for the 10th won't fire until all the previous 9 navigations and respondWith()'s have completed. That makes the start time stamp very low by comparison to the user perceived time to get to that destination, which is "Click 10th time" until "URL Change".

I'm not sure the use case for currently defined "navigation event start time" given the arbitrarily long start delay before it's recorded.

domenic:
Closes #8 by renaming. Closes #12 by changing the semantics to when the navigation is initiated.

This might be refined further as we figure out navigation queuing.

/cc @esprehn @mmocny @yoavweiss @igrigorik

I do worry a bit about the name `navigationStart`, as apparently in the [navigation timing API definition](https://w3c.github.io/navigation-timing/#dom-performancetiming-navigationstart), that is relative to the global monotonic clock, and not comparable to the values returned by `event.timeStamp` and `performance.now()`? So I worry this is a misleading symmetry. WDYT?

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,12 +550,14 @@ This can be useful for cleaning up any information in secondary stores, such as 
 
 The `window.appHistory` object has an event, `currententrychange`, which allows the application to react to any updates to the `appHistory.currentEntry` property. This includes both navigations that change its value, and calls to `appHistory.updateCurrentEntry()` that change its `state` or `url` properties. This cannot be intercepted or canceled, as it occurs after the navigation has already happened; it's just an after-the-fact notification.
 
-This event has one special property, `event.navigationStartTimeStamp`, which for same-document navigations gives the value of `event.timeStamp` for the corresponding `navigate` event. This allows it to be used for determining the overall load time, including the time it took for a promise passed to `e.respondWith()` to settle:
+This event has one special property, `event.navigationStart`, which for [same-document](#appendix-types-of-navigations) navigations gives the value of `performance.now()` when the navigation was initiated. This includes for navigations that were originally [cross-document](#appendix-types-of-navigations), like the user clicking on `<a href="https://example.com/another-page">`, but were transformed into same-document navigations by [navigation interception](#navigation-monitoring-and-interception). For completely cross-document navigations, `navigationStart` will be `null`.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -550,12 +550,14 @@ This can be useful for cleaning up any information in secondary stores, such as
...
 
-This event has one special property, `event.navigationStartTimeStamp`, which for same-document navigations gives the value of `event.timeStamp` for the corresponding `navigate` event. This allows it to be used for determining the overall load time, including the time it took for a promise passed to `e.respondWith()` to settle:
+This event has one special property, `event.startTime`, which for [same-document](#appendix-types-of-navigations) navigations gives the value of `performance.now()` when the navigation was initiated. This includes for navigations that were originally [cross-document](#appendix-types-of-navigations), like the user clicking on `<a href="https://example.com/another-page">`, but were transformed into same-document navigations by [navigation interception](#navigation-monitoring-and-interception). For completely cross-document navigations, `startTime` will be `null`.
```
yoavweiss, line 553: s/includes for navigations/includes navigations/?

README.md
```diff
@@ -565,7 +567,9 @@ appHistory.addEventListener("currententrychange", e => {
 });
 ```
 
-_TODO: could we populate this for cross-document navigations too? Then it kind of overlaps with existing timing APIs, and is probably a lot harder to implement..._
+_TODO: reconsider cross-document navigations. There will only be one (the initial load of the page); should we even fire this event in that case? (That's [#31](https://github.com/WICG/app-history/issues/31).) Could we give `startTime` a useful value there, if we do?_
+
+_TODO: is this property-on-the-event design the right one, or should we integrate with the performance timeline APIs, or...? Discuss in [#33](https://github.com/WICG/app-history/issues/33)._
```
yoavweiss, line 572: We should totally integrate this with the performance APIs, but that doesn't necessarily mean this timestamp is not needed. Are there other use cases for this timestamp, other than reporting? 
e.g. I know that we have [`Event.timestamp`](https://developer.mozilla.org/en-US/docs/Web/API/Event/timeStamp), which is awfully similar.

yoavweiss (commented):
Retroactive and belated LGTM

Issue #25, opened 2021-02-05
Include accessibility in the spec


MelSumner:
It would be good if the known needs of assistive technology were mentioned in this API, if only just to ensure that it isn’t forgotten or ignored in the inevitable edits & changes. 

Thanks for considering!

domenic:
For tracking, a lot of the background is discussed here: https://github.com/w3c/aria/issues/1353

I'm going to work on a PR to add some initial accessibility considerations right now, and will look forward to your thoughts.

domenic:
To give a quick status update on the work done so far and what remains:

- There seem to be two big problems with SPA navs and accessibility that people mention over and over: telling the AT user a navigation happened, and focus management.

- We believe telling the AT user that navigation happened slots really well into this API, with the `navigate` event and its `respondWith()` method. That is, this API has a first-class notion (much more so than `history.pushState()`) of telling the browser about a navigation and when it starts and ends, which the browser can then use in turn to tell AT about the navigation.

  On the Chromium side, where we've been prototyping the `navigate` event already, we've [investigated how to implement this](https://groups.google.com/a/chromium.org/g/chromium-accessibility/c/wDUwToU8LdM) and want to work on prototyping the AT integration soon as well.

- Focus management is a bit trickier. Although the app history APIs first-class notion of navigations helps here as well, there are conflicting opinions on what exactly should be done with focus on a navigation. Some developers we've talked to said that they prefer the cross-document behavior, where focus just resets to the body element, whereas others (backed by [research](https://www.gatsbyjs.com/blog/2019-07-11-user-testing-accessible-client-routing/)) indicate that something more app-specific would be best.

  Maybe by default the focus should be reset to the body, but there should be an easy way for a SPA router to make a more intelligent decision? This intelligent override could either be done using app history somehow, or with a new API (like `setSequentialFocusNavigationStartingPoint()`, which would be independently useful) that people use together with app history.

domenic:
I've posted a detailed proposal for focus management during SPA navs at https://github.com/WICG/app-history/issues/190. Thoughts welcome!

PR #28, opened 2021-02-05
Add initial discussion of accessibility goals and benefits

domenic:
Part of #25.

/cc @MelSumner @marcysutton @Robdel12

Next up for me is to read and digest https://www.gatsbyjs.com/blog/2019-07-11-user-testing-accessible-client-routing/...


MelSumner (commented):
Looks good. Thank you for adding this!

Issue #16, opened 2021-02-03
Handling failed navigations


domenic:
There are a few cases where navigations can fail:

1. `appHistory.back()` when the current entry is the 0th entry
2. `appHistory.forward()` when the current entry is the last entry
3. `appHistory.navigateTo(someKey)` where `someKey` is not the key of any entry in the list (e.g., it's one that's been permanently evicted, as discussed in [this section](https://github.com/WICG/app-history#per-entry-events); or, it's just a random string like `"not-a-key"`)

Currently the proposal indicates that (1)-(3) will have the promise immediately fulfill with `undefined`, in a way that's indistinguishable from a success. The proposal [explains](https://github.com/WICG/app-history#performing-navigations) you can check after the fact as to whether the navigation actually worked:

```js
const startingEntry = appHistory.currentEntry;
await appHistory.back();
if (startingEntry === appHistory.currentEntry) {
  console.log("We weren't able to go back, because there was nothing previous in the app history list");
}
```

It's unclear whether the current proposal gives a good design. I think at least (3) should result in a rejected promise.

I'm less sure about (1) or (2). Possibilities include:

- Current proposal (always fulfill promise with undefined)
- Fulfill the promise with a boolean indicating whether or not navigation succeeded
- Reject the promise if navigation failed. (This would be more palatable if we introduce the `canGoBack` and `canGoForward` sugar from #6.)

I don't have a strong opinion on which of these is best.

SetTrend:
In continuation of our discussion on [Navigation through the app history list](https://github.com/WICG/app-history/blob/main/README.md#navigation-through-the-app-history-list), let me suggest an `enum` to provde programmers a with a failure reason:
<br/>

```ts
const enum NavigateFailureReason
{
  NoError    // entry was set successfully
,  Error   // navigate() threw an exception
,  Cancelled    // navigate() was cancelled
,  InvalidHistoryEntry    // non-existend key
,  OutOfHistory    // .back()/forward() beyond entry array
}
```

tbondwilkinson:
I like 3) here - I think the Promise should reject in any case where the navigation did NOT occur as intended. So if your back is not possible, it didn't happen, and the Promise should reject.

I do think that the rejection reason should be helpful and tell the caller WHY the navigation didn't occur (cancelled by the router, not possible given the stack, etc.)

domenic:
[...]

PR #29, opened 2021-02-05
Clarify some joint session history relations

### LINKED ISSUE #23: Mention that pushNewEntry() also clears future entries
Both @natechapin and @jakearchibald have asked about this, so it's probably better to be explicit.

### LINKED ISSUE #24: Point out cases where appHistory.back()/forward() doesn't match history.back()/forward()
https://github.com/WICG/app-history#performing-navigations discusses how it only impacts your frame. But another important case is that history.back() can take you cross-origin, whereas appHistory.back() cannot.

domenic:
Closes #23. Closes #24.

Note that the failed-back detection might change per #16.

/cc @jakearchibald @natechapin

### DIFF
--- a/README.md
+++ b/README.md
@@ -742,7 +742,7 @@ This is done through:
 
 An `AppHistoryEntry` corresponds directly to a [session history entry](https://html.spec.whatwg.org/#session-history-entry) from the existing HTML specification. However, not every session history entry would have a corresponding `AppHistoryEntry` in a given `Window`: `AppHistoryEntry` objects only exist for session history entries which are same-origin to the current one, and contiguous.
 
-Example: if a browsing context contains session history entries with the URLs
+Example: if a browsing session contains session history entries with the URLs
 
 ```
 1. https://example.com/foo
@@ -773,7 +773,7 @@ _TODO: actually, we should probably expose scroll restoration mode, like `histor
 
 The view of history which the user sees, and which is traversable with existing APIs like `history.go()`, is the joint session history.
 
-Unlike the view of history presented by `window.history`, `window.appHistory` only gives a view onto session history entries for the current browsing context. This view does not present the joint session history, i.e. it is not impacted by frames. Notably, this means `appHistory.entries.length` is likely to be quite different from `history.length`.
+Unlike the view of history presented by `window.history`, `window.appHistory` only gives a view onto session history entries for the current browsing session. This view does not present the joint session history, i.e. it is not impacted by frames. Notably, this means `appHistory.entries.length` is likely to be quite different from `history.length`.
 
 Example: consider the following setup.
 
@@ -813,7 +813,7 @@ Traversal operates on the joint session history, which means that it's possible 
 
 - If code in the outer frame calls `appHistory.back()`, this will take us back to O2, and thus take the joint session history back to B. This means the inner frame will be navigated from `/inner-end` to `/inner-start`, changing its current app history entry from I2 to I1.
 
-- If code in the inner frame calls `appHistory.back()`, this will take us back to I1, and take the joint session history back to C. (This does not impact the outer frame.)
+- If code in the inner frame calls `appHistory.back()`, this will take us back to I1, and take the joint session history back to C. (This does not impact the outer frame.) The rule here for choosing C, instead of B, is that it moves the joint session history the fewest number of steps necessary to make I1 active.
 
 - If code in either the inner frame or the outer frame calls `history.back()`, this will take the joint session history back to C, and thus update the inner frame's current app history entry from I2 to I1. (There is no impact on the outer frame.)
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 The web's existing [history API](https://developer.mozilla.org/en-US/docs/Web/API/History) is problematic for a number of reasons, which makes it hard to use for web applications. This proposal introduces a new one, which is more directly usable by web application developers to address the use cases they have for history introspection, mutation, and observation/interception.
 
-This new `window.appHistory` API layers on top of the existing API and specification infrastructure, with well-defined interaction points. The main differences are that it is scoped to the current origin and frame, and it is designed to be pleasant to use instead of being a historical accident with many sharp edges.
+This new `window.appHistory` API [layers](#integration-with-the-existing-history-api-and-spec) on top of the existing API and specification infrastructure, with well-defined interaction points. The main differences are that it is scoped to the current origin and frame, and it is designed to be pleasant to use instead of being a historical accident with many sharp edges.
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
@@ -33,6 +33,9 @@ This new `window.appHistory` API layers on top of the existing API and specifica
   - [Introspecting the history list](#introspecting-the-history-list)
   - [Watching for navigations](#watching-for-navigations)
 - [Integration with the existing history API and spec](#integration-with-the-existing-history-api-and-spec)
+  - [Correspondence with session history entries](#correspondence-with-session-history-entries)
+  - [Correspondence with the joint session history](#correspondence-with-the-joint-session-history)
+  - [Integration with navigation](#integration-with-navigation)
 - [Impact on the back button and user agent UI](#impact-on-the-back-button-and-user-agent-ui)
 - [Security and privacy considerations](#security-and-privacy-considerations)
 - [Stakeholder feedback](#stakeholder-feedback)
@@ -197,9 +200,7 @@ The way for an application to navigate through the app history list is using `ap
 
 _TODO: realistic example of when you'd use this._
 
-Unlike the existing history API's `history.go()` method, which navigates by offset, navigating by key allows the application to not care about intermediate history entries; it just specifies its desired destination entry.
-
-There are also convenience methods, `appHistory.back()` and `appHistory.forward()`. Note that these only navigate through the _app history_ list, not the _joint session history_; that is, they do not navigate frames, and they cannot go back or forward to a cross-origin destination.
+Unlike the existing history API's `history.go()` method, which navigates by offset, navigating by key allows the application to not care about intermediate history entries; it just specifies its desired destination entry. There are also convenience methods, `appHistory.back()` and `appHistory.forward()`.
 
 All of these methods return promises, because navigations can be intercepted and made asynchronous by the `navigate` event handlers that we're about to describe in the next section. There are then several possible outcomes:
 
@@ -212,6 +213,8 @@ All of these methods return promises, because navigations can be intercepted and
 - The navigation succeeds, and was a same-document navigation. Then the promise fulfills with `undefined`,  and `appHistory.currentEntry` (as well as the URL bar, if appropriate) will update.
 
 - The navigation succeeds, and it was a different-document navigation. Then the promise will never settle, because the entire document and all its promises will disappear.
+
+As discussed in more detail in the section on [integration with the existing history API and spec](#integration-with-the-existing-history-api-and-spec), navigating through the app history list does navigate through the joint session history. This means it _can_ impact other frames on the page. It's just that, unlike `history.back()` and friends, such other-frame navigations always happen as a side effect of navigating your own frame; they are never the sole result of an app history traversal.
 
 ### Navigation monitoring and interception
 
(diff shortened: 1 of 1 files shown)


natechapin (approved):

README.md
```diff
@@ -193,7 +195,7 @@ _TODO: realistic example of when you'd use this._
 
 Unlike the existing history API's `history.go()` method, which navigates by offset, navigating by key allows the application to not care about intermediate history entries; it just specifies its desired destination entry.
 
-There are also convenience methods, `appHistory.back()` and `appHistory.forward()`.
+There are also convenience methods, `appHistory.back()` and `appHistory.forward()`. Note that these only navigate through the _app history_ list, not the _joint session history_; that is, they do not navigate frames, and they cannot go back or forward to a cross-origin destination.
```
jakearchibald, line 198: Is that possible? What if:

1. PageA contains an iframe to PageB.
1. `appHistory.pushNewEntry(…)` in top level page.
1. Navigate iframe PageB to PageC.

Now we have a joint history with 3 items, and we're currently on the third of those items.

What does `appHistory.entries` look like? What happens if `appHistory.back()` is called? And assuming the first entry in `appHistory.entries` points to step 1 above, what happens if it's traversed to? Does it also navigate the iframe?

domenic: Yeah, OK, I worked through this and it looks like you can navigate child frames. I don't think child frames can use app history to navigate parent frames...

I'll push a commit showing what I ended up with.

domenic:
Hmm, no, not ready to make a call here. Let's pull the discussion at https://github.com/WICG/app-history/pull/29/files#r571439026 out to top level. Here's what that scenario looks like in the current HTML spec:

```
// Outer frame session history
O1. https://example.com/page-a
O2. https://example.com/page-a-pushed

// Inner frame session history
I1. https://example.com/page-b
I2. https://example.com/page-c

// Joint session history: O1, I1, I2 / currently on I2

// App history in top-level
1. https://example.com/page-a         (associated to O1)
2. https://example.com/page-a-pushed  (associated to O2)

// App history in frame
1. https://example.com/page-b         (associated to I1)
2. https://example.com/page-c         (associated to I2)
```

This shows you what `appHistory.entries` looks like in each case.

> What happens if appHistory.back() is called?

If it's called from the top level, it updates the "current" pointer for the outer frame's session history to O1. Thus, we navigate the top level back to page-a, but leaves the iframe on page-c. That matches the behavior of  https://boom-bath.glitch.me/joint-sh-1.html if you click "pushState(), Navigate frame, history.go(-2)", at least in Chrome and Firefox.

The resulting joint session history is O1, O2, I1 / currently on O1.

If appHistory.back() is called from inside the frame, then similarly, we change the "current" pointer for the inner frame's session history to I1. That matches the behavior of  https://boom-bath.glitch.me/joint-sh-1.html if you click "pushState(), Navigate frame, history.back()", at least in Chrome and Firefox.

The resulting joint session history is O1, I1, I2 / currently on I1.

Can all this be made to work with the model at https://github.com/whatwg/html/issues/5767 ?

jakearchibald:
I'm assuming:

```
// Joint session history: O1, I1, I2 / currently on I2
```

…Is supposed to be

```
// Joint session history: O1, O2, I1, I2 / currently on I2
```

> If (`appHistory.back` is) called from the top level, it updates the "current" pointer for the outer frame's session history to O1. Thus, we navigate the top level back to page-a, but leaves the iframe on page-c. That matches the behavior of https://boom-bath.glitch.me/joint-sh-1.html if you click "pushState(), Navigate frame, history.go(-2)", at least in Chrome and Firefox.

This happens [due to a bug](https://source.chromium.org/chromium/chromium/src/+/master:content/browser/renderer_host/navigation_controller_impl.cc;drc=2ac64302ae161cd6b5e4b1254497bdf5fd6d3415;l=2974) (in Chrome at least, but I don't think it's intentional in other browsers either). I'd rather we fixed the bug than have to spec the bug.

https://iframe-session-history.glitch.me/ - if you navigate one iframe, then another, go(-2) will navigate both iframes back. The "if you navigate the parent then don't navigate the children at the same time" bug is super weird. I guess navigating back more than one entry at once is rare, so folks haven't really encountered it much, but `appHistory` will cause it to happen more often.

https://github.com/whatwg/html/issues/1191#issuecomment-776076679 - there's a possibility that the reason Chrome kept the bug is no longer an issue, but I'm not sure.

Oh, here's a better reason why we shouldn't spec that bug. Imagine a page with an iframe, and the following history:

1. Page at #, iframe at #
2. Iframe to #foo
3. Top page to #bar
4. Iframe to #hello - current step

`appHistory.back()` in the iframe will take us back to step 2, but due to the bug it will navigate the top page back to # and leave the iframe at #hello, which seems baaaaaaad. I'm pretty certain the correct behaviour is to take the page back to # and the iframe to #foo.

Another case to think about:

1. Page at #, iframe at #
2. Iframe to #foo
3. Top page to #bar - current step

What happens if the _iframe_ calls `appHistory.navigateTo(appHistory.currentEntry)`? This could either no-op, because it's already on the current entry, or it could move to step 2 of the joint session history, which would navigate the parent page back to #.

domenic:
> …Is supposed to be

I don't think so? We remove all current session history entries except the current entry of the joint session history, so we remove O2.

> This happens due to a bug (in Chrome at least, but I don't think it's intentional in other browsers either). I'd rather we fixed the bug than have to spec the bug.

Oh, I didn't realize this case was the one we were discussing in the other threads.

Changing this seems really unrealistic to me, given the compat implications surrounding the Gecko and Chrome code, and the fact that in general moving from interop to a new state is a hard sell. Plus, this behavior does what I want, and what I think authors generally want---it gives the ability to control your frame's history independent of any subframes or parent frames!

> Oh, here's a better reason why we shouldn't spec that bug. Imagine a page with an iframe, and the following history:
>
> [...]
>
> appHistory.back() in the iframe will take us back to step 2, but due to the bug...

Well... I'm _hoping_ we can avoid thinking of appHistory as moving through the joint list, and instead just talk about it navigating the individual frame. So, what it does, is somehow *waves hands* moves the iframe back to #foo, while leaving the top page untouched.

Do you think there's any realistic model of history that is browser-compatible while allowing operations like that? In particular, app history would really benefit from a local this-browsing-context-only session history, like the current spec has.

> Another case to think about:
>
> [...]
>
> What happens if the iframe calls appHistory.navigateTo(appHistory.currentEntry)? 

It should be a no-op.

domenic:
Re-reading your response and my response to it, I think we may have over-focused on the particular example and its related bug. My larger point and question was the following:

- The current spec has a model, where there are independent per-browsing context session histories, each with their own "current" pointer.

- This model would be extremely useful for meeting the current design vision of app history.

- We have some evidence that cases which you wouldn't expect in a pure joint SH model, but would only be possible with an independent per-BC session histories, can occur. (My demo/the bug.) So a per-BC model is probably not un-implementable, and might even match implementations?

- That said, we really need a stable and sensible foundation for session history, browsing sessions, and the like, before we introduce anything new like app history. Your current work is our best hope at that. So the question is, do you see a path forward for getting this kind of per-BC behavior, while still rationalizing session history according to your plan (or some slight tweak to it)? Or, is your work purely based on a joint session history model that would clash badly with allowing independently varying per-BC current pointers?

domenic:
After some offline discussions, we're going to proceed with a joint session history model for now. This means app history *can* impact other frames (although it doesn't give a *view* into them).

One major reason of this is that, if appHistory can move a frame through its frame-specific session history independently, the developer can put the user into a state they've never seen before. This then raises weird questions like, what does the back button do? What does the forward button do? And, of course, there's the potential implementation and spec complexity.

So I'll update this PR in that direction. Separately, @jakearchibald seems to be working on a sort of library of crazy nested iframe cases. When that's available, we'll want to analyze each of them in terms of how appHistory operations and views behave.

domenic:
Alright, this has been updated. I'd love @natechapin and @jakearchibald to take another look before merging. Also, if he has time, @creis's perspective would be much appreciated.

jakearchibald:
> Separately, @jakearchibald seems to be working on a sort of library of crazy nested iframe cases. When that's available, we'll want to analyze each of them in terms of how appHistory operations and views behave.

haha err it's more of a little presentation on what I'm trying to do with the session history spec. It does cover a few history cases the current spec doesn't handle, but I'm not sure how useful that'll be here.

README.md
```diff
@@ -752,38 +769,67 @@ Apart from these new fields, the session history entries which correspond to `Ap
...
-
-## Impact on the back button and user agent UI
+The view of history which the user sees, and which is traversable with existing APIs like `history.go()`, is the joint session history.
 
-The app history API doesn't change anything about how user agents implement their UI: it's really about developer-facing affordances. Users still care about the joint session history, and so that will continue to be presented in UI surfaces like holding down the back button. Similarly, pressing the back button will continue to navigate through the joint session history, potentially across origins and out of the current app history (into a new app history, on the new origin).
+Unlike the view of history presented by `window.history`, `window.appHistory` only gives a view onto session history entries for the current browsing context. This view does not present the joint session history, i.e. it is not impacted by frames. Notably, this means `appHistory.entries.length` is likely to be quite different from `history.length`.
```
jakearchibald, line 776: It seems fine to expose entries that cross browsing contexts via COOP+COEP

README.md
```diff
@@ -752,38 +769,67 @@ Apart from these new fields, the session history entries which correspond to `Ap
...
 
-An important consequence of this is that when iframes are involved, the back button may navigate through the joint session history, without changing the current _app history_ entry. For example, consider the following sequence:
+Example: consider the following setup.
 
-1. `https://example.com/start` loads
+1. `https://example.com/start` loads.
 1. The user navigates to `https://example.com/outer` by clicking a link. This page contains an iframe with `https://example.com/inner-start`.
+1. Code on `https://example.com/outer` calls `appHistory.pushNewEntry({ url: "/outer-pushed" })`
```
jakearchibald, line 782:
```suggestion
1. Code on `https://example.com/outer` calls `appHistory.pushNewEntry({ url: "/outer-pushed" })`.
```

README.md
```diff
@@ -752,38 +769,67 @@ Apart from these new fields, the session history entries which correspond to `Ap
...
+```
+O1. https://example.com/start        (associated to A)
+O2. https://example.com/outer        (associated to B)
+O3. https://example.com/outer-pushed (associated to C and D)
+```
+
+The app history list for the inner frame looks like:
+
+```
+I1. https://example.com/inner-start  (associated to C)
```
jakearchibald, line 808:
```suggestion
I1. https://example.com/inner-start  (associated to B and C)
```

README.md
```diff
@@ -752,38 +769,67 @@ Apart from these new fields, the session history entries which correspond to `Ap
...
+I2. https://example.com/inner-end    (associated to D)
+```
+
+Traversal operates on the joint session history, which means that it's possible to impact other frames. Continuing with our previous setup, and assuming the current entry in the joint session history is D, then
```
jakearchibald, line 812:
```suggestion
Traversal operates on the joint session history, which means that it's possible to impact other frames. Continuing with our previous setup, and assuming the current entry in the joint session history is D, then:
```

README.md
```diff
@@ -752,38 +769,67 @@ Apart from these new fields, the session history entries which correspond to `Ap
...
+
+- If code in the outer frame calls `appHistory.back()`, this will take us back to O2, and thus take the joint session history back to B. This means the inner frame will be navigated from `/inner-end` to `/inner-start`, changing its current app history entry from I2 to I1.
+
+- If code in the inner frame calls `appHistory.back()`, this will take us back to I1, and take the joint session history back to C. (This does not impact the outer frame.)
+
+- If code in either the inner frame or the outer frame calls `history.back()`, this will take the joint session history back to C, and thus update the inner frame's current app history entry from I2 to I1. (There is no impact on the outer frame.)
```
jakearchibald, line 818: This is an interesting outcome, because `C` isn't an entry in the iframe, it belongs to the parent.

I'm not saying it's wrong, but it's something that might need further documentation.

You could say: To traverse to a specific `appHistory` entry, joint history should traverse the least number of steps to make that history entry active.

That means you'd end up on a different item of joint session history depending on the direction of travel. I can't decide if that's intuitive or not.

The alternative is to _always_ travel to the step in joint history where that entry was created. That means it's consistent going back & forward, but may result in changing other frames more than necessary. In this example it would navigate the parent frame (and of course hit that fun bug).

jakearchibald (approved):
Couple of corrections & questions, but happy for this to land and the questions broken out into separate issues. Enjoying where this is heading!

README.md
```diff
@@ -723,18 +730,28 @@ The app history API provides several replacements that subsume these events:
...
-1. `https://example.com/foo`
-1. `https://example.com/bar`
-1. `https://other.example.com/whatever`
-1. `https://example.com/baz`
+- Ensuring that traversal through app history always maps to a traversal through the joint session history, i.e. a traversal which is already possible today.
 
-then, if the current entry is (4), there would only be one `AppHistoryEntry` in `appHistory.entries`, corresponding to (4) itself. If the current entry is (2), then there would be two `AppHistoryEntries` in `appHistory.entries`, corresponding to (1) and (2).
+### Correspondence with session history entries
+
+An `AppHistoryEntry` corresponds directly to a [session history entry](https://html.spec.whatwg.org/#session-history-entry) from the existing HTML specification. However, not every session history entry would have a corresponding `AppHistoryEntry` in a given `Window`: `AppHistoryEntry` objects only exist for session history entries which are same-origin to the current one, and contiguous.
```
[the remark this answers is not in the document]
domenic: It does interact, indeed. However, the _app history state_ gets copied over (shared? not yet clear) for fragment navigations; this is a semantic difference between "app history state" and "serialized state", which is discussed in this section a bit lower.

README.md
domenic: Yeah, I'll tweak this to "browsing session" I guess? It's not precise.

README.md
domenic: Yeah, I was going for least number of steps... hmm. I'll try to capture this and maybe open a tracking issue.

domenic: Although I assume your comment is meant to apply to the second bullet, not the third. The third uses `history.back()`, not `appHistory.back()`.

jakearchibald: Ah yes sorry

natechapin (approved):

PR #46, opened 2021-02-25
Make all same-document navigations sync

### LINKED ISSUE #19: Worries about making all navigations async
The proposal currently makes all navigations asynchronous. This is to support allowing them to be intercepted by `navigate` and replaced by SPA navs. However, the implications of this are a bit tricky.

Note that today, at least in Firefox and Chrome,

```js
location.hash = "foo";
console.log(location.href);
```

outputs the new URL, with `#foo` at the end. `history.pushState()` and `history.replaceState()` are [a non-interoperable mess](https://docs.google.com/document/d/1Pdve-DJ1JCGilj9Yqf5HxRJyBKSel5owgOvUJqTauwU/edit#), but at least some of the time they're also synchronous. `aElement.click()` for a fragment navigation link is [asynchronous in Firefox, synchronous in Chrome](https://jsbin.com/suhodidawi/edit?html,console,output), hmm.

But the upside is that if we stick with the current proposal that `location.href` only updates at the "end" of the navigation, making navigations async this would be a breaking change.

There are a few potential mitigations I can think of, although none of them are great:

- Allow `location.href` (and other things, like `document.URL`) to update, but still treat the navigation _process_ as async, potentially reverting the URL if the navigation gets canceled. This seems very messy and likely to lead to bugs and problems, both in application code and browser code.

- Only apply this new navigations-are-async behavior if an event listener is present for `appHistory`'s `navigate` event. Adding event listeners should not have side effects, so we'd probably change the API shape to something like `appHistory.setNavigateHandler()` instead of using an event. This seems doable, although perhaps intrusive into browser code.

- Only make navigations async for cases where they are already async in at least one browser, such as:
  - New APIs like `appHistory.pushNewEntry()`
  - Cross-document navigations (including `aElement.click()` for non-fragment navigations)
  - `aElement.click()` for fragment navigations, on the theory that other browsers can align with Firefox?
  - _NOT_ `location.hash` or `location.href` updates that only change the hash
  - _NOT_ `history.pushState()` (probably; pending further browser testing)

I'm not sure what the right path is here. It might be a more radical re-think, beyond these options. I also need to process the issues @esprehn filed around how queued-up navigations are problematic as currently proposed (#10 #11 #13); it's possible the process of solving those might change how we approach this problem.

### LINKED ISSUE #44: appHistory.update async or not, and the effect on navigation cancellation
In the example section there is 

```js
// Updates the URL shown in the address bar, as well as the url property. `state` stays the same.
appHistory.update({ url });

// You can also explicitly null out the state, instead of carrying it over:
appHistory.update({ url, state: null });

// Only updates the state property.
appHistory.update({ state });

// Update both at once.
appHistory.update({ url, state });
```

which seems to indicate that it's not async. There is also the WebIDL format:

```
undefined update(optional AppHistoryEntryOptions options = {});
undefined update(AppHistoryNavigationCallback);
```

However, there is one place where it is written with `await`

> use `await appHistory.update({ state, url })`

But the real reason for bringing this up is to understand better how `appHistory.update` and the `navigate` event are intended to work together: 

If `update` is sync, then does that mean `event.respondWith()` has no effect and you cannot cancel an `update` with it? Would `navigate` event handlers need to differentiate between `push` and `update` in those cases? Or would it still work and just roll back the `update` at some future point, possibly firing the `navigate` event again?

Also, if `update` and `push` are different, does that open the door to potential mixups by developers mis-remembering which needs to be `await`-ed or not? (specifically regarding [this section](https://github.com/WICG/app-history/blob/main/README.md#navigations-while-a-navigation-is-ongoing)?)

Hopefully my questions make sense, and also I hope I didn't miss a section in the docs that explain this. 😃 Thanks!

(edit: This probably ties somewhat into the conversation happening in #19 as well)

domenic:
This includes ones generated by intercepting the navigate event with event.respondWith(). However, the promise passed to event.respondWith() is still useful for signaling the navigation's successful or unsuccessful completion. That is used to fuel browser UI, such as the loading spinner, as well as promises and events. And a rejected promise passed to event.respondWith() will roll back the navigation.

Closes #19. Closes #44.

### DIFF
--- a/interception-details.md
+++ b/interception-details.md
@@ -107,6 +107,8 @@ After the promise fulfills in ten seconds:
 
 ### Delayed failure
 
+**The behavior here is being debated in [#47](https://github.com/WICG/app-history/issues/47) and is not finalized.**
+
 ```js
 appHistory.addEventListener("navigate", e => {
   e.respondWith(new Promise((r, reject) => setTimeout(() => reject(new Error("bad")), 10_000)));
@@ -146,6 +148,8 @@ After the promise rejects in ten seconds:
 1. The no-longer current `AppHistoryEntry` representing `/foo` fires `dispose`.
 1. Any loading spinner UI stops.
 1. The second `console.log()` outputs `""` (or whatever the old URL was).
+
+Note: any unreachable `AppHistoryEntry`s disposed as part of the synchronous block do not get resurrected.
 
 ### Interrupting a slow navigation
 
--- a/interception-details.md
+++ b/interception-details.md
@@ -1,0 +1,215 @@
+# App history navigation details
+
+[Navigation interception](./README.md#navigation-monitoring-and-interception) using app history can be complicated. There are a lot of moving parts and events. This document details several scenarios and walks through how they work in detail.
+
+TODO: research where in these sequences accessibility technology should integrate, by researching where AT integrates into typical cross-document navigations.
+
+## Same-document
+
+### Immediate successful `respondWith()`
+
+```js
+appHistory.addEventListener("navigate", e => {
+  e.respondWith(Promise.resolve());
+});
+
+location.hash = "#foo";
+console.log(location.hash);
+```
+
+Synchronously:
+
+1. `navigate` fires on `window.appHistory`.
+1. `appHistory.current` fires `navigatefrom`.
+1. `location.href` updates.
+1. `appHistory.current` updates. `appHistory.current.finished` is `false`.
+1. `currentchange` fires on `window.appHistory`.
+1. `appHistory.current` fires `navigateto`.
+1. Any now-unreachable `AppHistoryEntry` instances fire `dispose`.
+1. The `console.log()` outputs `"#foo"`
+
+Asynchronously but basically immediately:
+
+1. The URL bar updates.
+
+After the promise settles in one microtask:
+
+1. `appHistory.current.finished` changes to `true`.
+1. `appHistory.current` fires `finish`.
+1. `navigatefinish` is fired on `appHistory`.
+
+### No interception
+
+```js
+appHistory.addEventListener("navigate", e => {
+  // Do nothing.
+});
+
+location.hash = "#foo";
+console.log(location.hash);
+```
+
+This is the same as the previous case.
+
+### Cancelation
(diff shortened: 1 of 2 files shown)


interception-details.md
```diff
@@ -0,0 +1,215 @@
...
+
+Synchronously:
+
+1. `navigate` fires on `window.appHistory`.
+1. `appHistory.current` fires `navigatefrom`.
+1. `location.href` updates.
+1. `appHistory.current` updates. `appHistory.current.finished` is `false`.
+1. `currentchange` fires on `window.appHistory`.
+1. `appHistory.current` fires `navigateto`.
+1. Any now-unreachable `AppHistoryEntry` instances fire `dispose`.
```
frehner, line 131: for delayed failures, what happens to these "now-unreachable" Entries? should they become "reachable" again since the navigation was cancelled and `appHistory.current` was rolled back to its previous value?

interception-details.md
```diff
@@ -0,0 +1,215 @@
...
+1. The `console.log()` outputs `"/foo"`.
+
+Asynchronously but basically immediately:
+
+1. The URL bar updates to show `/foo`.
+1. Any loading spinner UI starts.
+
+After the promise rejects in ten seconds:
```
frehner, line 139: should there be an additional entry in the list below about the `appHistoryEvent.signal` being aborted?

frehner (commented):
I really like the introduction of `appHistoryEvent.signal` 👍

interception-details.md
domenic: Great question.

I'm inclined to say that they're gone forever? Splicing a future history back into existence seems pretty weird. But in general the introduction of fallible single-page navigations raises a lot of questions, hmm...

interception-details.md
domenic: I don't think so. The signal property is specifically for aborts. I.e., it's meant to indicate that something else aborted the navigation, not that the navigate handler completed the navigation with a failure.

interception-details.md
domenic: I'll open a dedicated issue about this and then point to it from here.

PR #48, opened 2021-02-26
Add a prominent summary section

domenic:
The explainer is very long! Hopefully this will let people figure out the basic idea without having to read the whole thing.

There's definitely some overlap with "Problem statement", but I think it's OK-ish... feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,26 @@ The web's existing [history API](https://developer.mozilla.org/en-US/docs/Web/AP
 
 This new `window.appHistory` API [layers](#integration-with-the-existing-history-api-and-spec) on top of the existing API and specification infrastructure, with well-defined interaction points. The main differences are that it is scoped to the current origin and frame, and it is designed to be pleasant to use instead of being a historical accident with many sharp edges.
 
+## Summary
+
+There's a lot going on here, and this explainer tries to really explain it! Here's the short version.
+
+The existing [history API](https://developer.mozilla.org/en-US/docs/Web/API/History) is bad, especially for single-page applications. This leads to user-facing pain in the form of lost state and broken back buttons, and developer-facing pain in the form of maintaining lots of hacky code to achieve the desired user experience.
+
+The proposed `window.appHistory` API solves this by providing the following:
+
+- Robust introspection and state tracking for your web application's entire history list. For example, instead of a single `history.state` value which disappears whenever the user performs a fragment navigation, and an unreliable `length` property, you can inspect the URL and state of any past or future history entry. Instead of [unreliable](https://github.com/whatwg/html/issues/5562) `popstate` and `hashchange` events, you get a `currentchange` event which fires on any change in the current history entry.
+
+- Built-in interception of all navigations, no matter how they are initiated. This means that instead of forcing all parts of your application through a router-library-supplied navigation function, or adding a global `click` handler to catch `<a>` clicks and convert them into single-page navigations, you can listen to a central `navigate` event and use that to update the DOM to perform a single-page app navigation. This `navigate` event gets all sorts of useful information, such as the destiniation URL and state, whether the navigation was user-initiated, any form data that was submitted, and more.
+
+- More ergonomic navigation methods. For example, instead of `history.pushState(undefined, undefined, url)`, you can use `appHistory.push({ url })`. And the latter will return a promise, to signal when the navigation completes! Similarly, instead of `history.go(n)`, you can use `appHistory.navigateTo(key)`, where `key` is a unique identifier that each app history entry has.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,26 @@ The web's existing [history API](https://developer.mozilla.org/en-US/docs/Web/AP
 
 This new `window.appHistory` API [layers](#integration-with-the-existing-history-api-and-spec) on top of the existing API and specification infrastructure, with well-defined interaction points. The main differences are that it is scoped to the current origin and frame, and it is designed to be pleasant to use instead of being a historical accident with many sharp edges.
 
+## Summary
+
+There's a lot going on here, and this explainer tries to really explain it! Here's the short version.
+
+The existing [history API](https://developer.mozilla.org/en-US/docs/Web/API/History) is bad, especially for single-page applications. This leads to user-facing pain in the form of lost state and broken back buttons, and developer-facing pain in the form of maintaining lots of hacky code to achieve the desired user experience.
```
chrishtr, line 11:
```suggestion
The existing [history API](https://developer.mozilla.org/en-US/docs/Web/API/History) is hard to deal with in practice, especially for single-page applications. This leads to user-facing pain in the form of lost state and broken back buttons, and developer-facing pain in the form of maintaining lots of hacky code to achieve the desired user experience.
```

chrishtr (commented):
My main feedback is I think the TL;DR is not TL;DR enough. I'd try to delete as much about the details as possible, and instead say something like:

Existing history API has some problems, leading in the best case to workarounds and hacks, and in the worst case to seemingly-unending broken corner cases.

The main problems are:

A. Looking back at the existing history stack, and storing data associated with history entries is fragile. This is because the data can be lost sometimes, and the browser sometimes inserts entries that are hard to understand due to various side-effects like iframes.
  We solved A with a way to view only states caused by the application, not the magic-browser ones, and with a way of storing data that cannot be lost.

B. It's hard to figure out all the ways navigations might occur, so that the application can override ones that need to stay on the site.
  We solved this by exposing events that allow the application to observe navigation actions, and prevent default behavior.

C. Some of the history APIs are pretty ugly or hard to understand.
  We solved this by making new interfaces that are easy for developers to understand and use.

And skip all details to the non-TL;DR. I also suggest following the TL;DR with an Example section that shows it in action. People like examples.

README.md
```diff
@@ -4,6 +4,26 @@ The web's existing [history API](https://developer.mozilla.org/en-US/docs/Web/AP
...
+
+A key reason for this new API, as opposed to extending the `window.history` API, is because we need to avoid `window.history`'s original sin: its over-broad scope. In particular, if an iframe navigates, this contributes to the user's view of session history, e.g. what happens when they press the back button. But for application code, this is not helpful: iframe navigations are generally independent of the application, so the fact that they can mess with session history makes `history.back()` and `history.state` unreliable. Similarly, because clicking back can cause the user to move across origins, `window.history` includes cross-origin information, which means we can't add better introspection facilities (such as the [previously-proposed `history.index`](https://github.com/whatwg/html/issues/2710)) without leaking cross-origin information. App history separates the user's view of history from the web application's, by giving a view onto the same-origin, same-frame, contiguous history entries only. This opens the door to allow in-depth introspection and manipulation of the history list.
+
```
tbondwilkinson, line 22: This one is just too wordy. It's detailed in a summary!

I think something like this instead:

- `appHistory` provides a same-origin, same-frame, contiguous history view of the browser history. History entries added by iframes are not visible through `appHistory`, and no events will be sent for those navigations. This fixes longstanding issues with iframes polluting the history stack and opens the door for in-depth introspection and manipulations of application owned history.

README.md
```diff
@@ -4,6 +4,26 @@ The web's existing [history API](https://developer.mozilla.org/en-US/docs/Web/AP
...
+Additionally, when single-page navigations are implemented by intercepting the `navigate` event, they become first-class citizens. Browser features like the loading spinner, or accessibility technology, can key off of this standardized entry point. And other APIs like performance timing APIs can measure the duration of such navigations.
```
tbondwilkinson, line 23:
```suggestion
`appHistory` will enable browsers to implement features like the loading spinner or accessibility technology for in-app navigations. And other APIs like performance timing APIs can measure the duration of such navigations more accurately.
```

domenic:
Thanks both! I've put up a new revision based largely on @chrishtr's feedback, but would love a re-review from you both.

README.md
```diff
@@ -4,6 +4,61 @@ The web's existing [history API](https://developer.mozilla.org/en-US/docs/Web/AP
...
+A page-supplied "back" button can actually take you back, even after reload, by inspecting the previous history entries:
+
+```js
+backButtonEl.addEventListener("click", () => {
+  if (appHistory.entries[appHistory.entries.current.index - 1]?.url === "/product-listing") {
+    // Go back to that entry
+    appHistory.back();
+  } else {
+    // Go "forward" to that entry, e.g. if the user arrived here by typing the URL directly.
+    appHistory.push("/product-listing");
```
tbondwilkinson, line 47: More likely this would be a replace?

tbondwilkinson (commented):
Looks good!

Issue #42, opened 2021-02-18
Should appHistory.push() and update() fire the navigate event?


dvoytenko:
> Programmatically-initiated same-document navigations initiated via appHistory.push(), appHistory.update(), or their old counterparts history.pushState() and history.replaceState().

The older `history.pushState` and counterpart did not navigate the document. This is used extensively by routers to reconcile navigations with the history structure when a state might need to be added or updated to provide for back behavior and other application features. There are already several ways to navigate, including `location = '...'`, `location.assign()`, `link.click()`, etc. Perhaps the navigation in this API is unnecessary and might be even harmful.

domenic:
It depends on what you mean. They definitely navigate the document in the sense of how the spec and implementation sees it. (They're same-document navigations, like a fragment navigation.) So I don't think it's possible to call them harmful in that way, since it's something history.pushState already does.

Could you expand on this a bit more? The intention is for appHistory.push() to navigate the document in the same way history.pushState() does, i.e. be a same-document navigation that updates the URL/state but does not update the `Document` object. Does that perhaps ameliorate your concern?

dvoytenko:
Hmm. I don't think `history.pushState` does this today. Could you clarify this? No navigation of any kind is triggered via that API as far as I know.

domenic:
We discussed this a bit offline. Whether or not pushState is a navigation aside, the main concern is about the `navigate` event integration.

In particular, today the model people use is: do stuff to update the page, and then use `history.pushState()` to update the URL. (Or perhaps in the opposite order, but the idea is the same: `history.pushState()`s job is to update the page's URL to reflect its contents.)

Whereas, the mental model of a pure app history API user is: perform a navigation, either with `location.href = ` or `history.pushState()` or `appHistory.push()` or.... Then, in the `navigate` handler, update the page.

@dvoytenko's is concerned about mixed codebases, e.g. ones migrating from the current model to the app history model. How does their `navigate` handler deal with the fact that in one part of their code, they're doing `updateUI(); history.pushState()`, and in other parts of their code, they're just doing `appHistory.push()` and expecting the `navigate` handler to call `updateUI()`?

One solution here, which I think is pretty good, is to be more explicit about the navigation initiator. E.g., inside the `navigate` event, we could have an `event.initiator` which is an enum discussing [what started the navigation](https://github.com/WICG/app-history#appendix-types-of-navigations). Then you could write code like

```js
appHistory.addEventListener("navigate", e => {
[...]

PR #54, opened 2021-03-03
Make push()/update() navigate, and change state interface

### LINKED ISSUE #36: App history state (im)mutability
Continuing from https://github.com/WICG/app-history/issues/7#issuecomment-777756310, where @tbondwilkinson notes a sharp edge of the existing `history.state` API:

```js
history.pushState({ test: 2 }, null, '');
console.log(history.state.test); // 2
history.state.test = 3;
console.log(history.state.test); // 3

// refresh the page

console.log(history.state.test); // 2
```

That is, because it allows any arbitrary JavaScript object, it allows mutable JavaScript objects. But the object is only serialized into (semi-)persistent storage when you call `history.pushState()` or `history.replaceState()`.

The current proposal duplicates this sharp edge with app history state. E.g.,

```js
appHistory.push({ state: { test: 2 } });
console.log(appHistory.current.state.test); // 2
appHistory.current.state.test = 3;
console.log(appHistory.current.state.test); // 3

// refresh the page

console.log(appHistory.current.state.test); // 2
```

This seems quite bad to me, and I think we should consider a way to fix it, if we can do so without interfering with other use cases. I'll post a reply with some suggestions.

domenic:
Closes #36 by moving from appHistoryEntry.state to getState() and setState() methods. This adds the ability to modify the state of non-current entries as a bonus.

Solves much of #42 by making push() and update() navigate by default, although it does not yet include the initiator idea.

As part of this we simplify update() significantly, since now if you want to change state without performing a navigation, you would use setState(). This makes the URL argument required for it. For symmetry, I removed url from the options for push().

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,24 +65,25 @@ appHistory.addEventListener("currentchange", e => {
 - [Problem statement](#problem-statement)
 - [Goals](#goals)
 - [Proposal](#proposal)
-  - [The current entry, and single-page navigations](#the-current-entry-and-single-page-navigations)
+  - [The current entry](#the-current-entry)
   - [Inspection of the app history list](#inspection-of-the-app-history-list)
   - [Navigation through the app history list](#navigation-through-the-app-history-list)
   - [Navigation monitoring and interception](#navigation-monitoring-and-interception)
+    - [Example: replacing navigations with single-page app navigations](#example-replacing-navigations-with-single-page-app-navigations)
     - [Accessibility benefits of standardized single-page navigations](#accessibility-benefits-of-standardized-single-page-navigations)
     - [Measuring standardized single-page navigations](#measuring-standardized-single-page-navigations)
-    - [Example: replacing navigations with single-page app navigations](#example-replacing-navigations-with-single-page-app-navigations)
     - [Example: single-page app "redirects"](#example-single-page-app-redirects)
     - [Example: cross-origin affiliate links](#example-cross-origin-affiliate-links)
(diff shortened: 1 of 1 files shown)


frehner:
This PR cleared up a misunderstanding I had, thanks!

And just to clarify, it appears that `appHistory.current.setState()` doesn't fire the navigate event, correct?

But I assume it would fire the `currentchange` event? 

And should it fire any event if you do `setState()` on an entry that isn't the current one?

domenic:
> And just to clarify, it appears that appHistory.current.setState() doesn't fire the navigate event, correct?

Correct.

> But I assume it would fire the currentchange event?
> 
> And should it fire any event if you do setState() on an entry that isn't the current one?

Good questions. I think it should not fire currentchange, but probably it should have its own per-entry event. Let me add that.

domenic:
I'm going to merge this now so I can build more on top of it, but any further reviews after the fact are appreciated!

tbondwilkinson:
I'm not sure about `setState()`.

I think it creates the same problem that we have today with `history.pushState()` where it's updating the browser about the current state of things AFTER that state has been changed. So on the one hand we're encouraging people to do navigations by storing information in history and responding to those data changes in the navigate event, but this new `setState()` method is an out - it basically means that there's a workaround to update things in history without ever triggering a navigate event (where we would expect most people's rendering code to be). I think that dilutes the model and is pretty niche.

Further, it means that some part of the application can just clear your state and you don't get any event about it besides currentchange. That also feels bad.

PR #55, opened 2021-03-03
Add entry.index and appHistory.canGoBack/canGoForward

### LINKED ISSUE #6: Adding sugar for current index/can go back/can go forward
(Moved from https://github.com/slightlyoff/history_api/issues/25 by @SetTrend, summarized by me.)

It's currently slightly awkward to determine the index of the current entry, and whether you can go back and forward:

```js
const index = appHistory.entries.indexOf(appHistory.currentEntry);
const canGoBack = index !== 0;
const canGoForward = index < appHistory.entries.length - 1;
```

It might be nice if this could be done with

```js
appHistory.currentIndex; // just index? or currentEntryIndex?
appHistory.canGoBack;
appHistory.canGoForward;
```

or maybe we should add something to all the individual entries, like

```js
appHistory.currentEntry.index;
appHistory.currentEntry.canGoBack; // isFirst?
appHistory.currentEntry.canGoForward; // isLast?
```

A use case where this is especially important is when you want to show your own navigation UI, such as when you're a widget inside an iframe (where the browser navigation UI is not very related to your UI). @SetTrend's specific example was a multi-step form embedded in an iframe.

domenic:
Closes #6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ The entry point for the app history API is `window.appHistory`. Let's start with
 
 - `key`: a user-agent-generated UUID identifying this history entry. In the past, applications have used the URL as such a key, but the URL is not guaranteed to be unique.
 
+- `index`: the index of this `AppHistoryEntry` within the app history list. (Or, `-1` if the entry is no longer in the list, or not yet in the list.)
+
 - `url`: the URL of this history entry (as a string).
 
 - `sameDocument`: a boolean indicating whether this entry is for the current document, or whether navigating to it will require a full navigation (either from the network, or from the browser's back/forward cache). Note: for `appHistory.current`, this will always be `true`.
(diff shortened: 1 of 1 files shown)


frehner:
Do you think there would be a desire to allow `navigateTo()` (or a similarly-named method) to take in the index? 

I think it's not a big deal if it doesn't (e.g. `appHistory.navigateTo(appHistory.entries[index].key)` ) 

(maybe somewhat related to the discussion in #7 as well)

domenic:
Interesting. I'm unsure... I do think the discussion in #7 is more grounded in real use cases. Indices are a bit fragile.

frehner:
> Do you think there would be a desire to allow `navigateTo()` (or a similarly-named method) to take in the index?
> 
> I think it's not a big deal if it doesn't though (e.g. `appHistory.navigateTo(appHistory.entries[index].key)` )
> 
> (maybe somewhat related to the discussion in #7 as well)

The more I think about this, the more I think it's not necessary - in any case where you have an entry's `index` you'll also have its `key` (right?). And like you mentioned, the indices are fragile.

Issue #32, opened 2021-02-07
Cancelling UI initiated navigations (back/forward)


posva:
Hello, looking at the part regarding navigation cancellation (https://github.com/WICG/app-history#navigation-monitoring-and-interception) and later on, the comparison with navigation guards makes me wonder why can't UI based navigations (or `history.back())` be cancelled as well with `event.preventDefault()` **for same document and origin** navigations? If they can't be cancelled navigation guards cannot be implemented like proposed since they also run when the user navigates backwards and forwards.
Currently, on Vue Router 4, I restore the history entry (when possible): https://github.com/vuejs/router/blob/main/packages/router/src/router.ts#L1039-L1042 when a ui initiated navigation is cancelled.

It would be even nicer if the URL stays the same until the navigation is confirmed, **no matter what initiated it**, making it consistent when doing `history.back()` (or clicking the ui button) or clicking on a link of the application. In applications, users show modals or confirmations messages when before navigating away from a page (e.g. to not lose unsaved changes) and the URL changing is often confusing

**All of the above is for same origin and document navigations**, which is the context of an SPA.

Yay295:
To prevent someone from trapping a user on their page by canceling all navigations.

posva:
> To prevent someone from trapping a user on their page by canceling all navigations.

I updated the description, what I said is only relevant for the same origin and document (e.g. SPA) navigations

domenic:
Hey, thanks for opening this! Unfortunately, we can't allow intercepting the URL bar or back/forward buttons, even for same-document navigations. Consider the following attack:

- The user visits your site
- You call `appHistory.pushNewState()` the first time they click a button, performing a same-document navigation.
- You now intercept the `navigate` event and always call `e.preventDefault()` or `e.respondWith()`.

You have now completely disabled the user's back button, since pressing back does a same-document navigation, but that navigation never succeeds.

This isn't an acceptable outcome for our users, so we can't allow interception in this way.

We _can_ allow interception of same-document _URL-bar_ triggered navigations, i.e. if the user changes the fragment component. I'm working to clarify that in #26. There's also no problem with intercepting `history.back()`, or clicking links; those are all interceptable. But for the browser UI's back/forward button, interception is too abuse-prone, even same-document. 

Does this make sense?

posva:
> We can allow interception of same-document URL-bar triggered navigations, i.e. if the user changes the fragment component.

Can we intercept it if they change the pathname or the search through the back and forward buttons?

[...]

PR #56, opened 2021-03-03
Revamp when navigate fires, is cancelable, and is respond-able

### LINKED ISSUE #51: Naming and precise definition of event.sameOrigin
For [the navigate event](https://github.com/WICG/app-history#navigation-monitoring-and-interception), the explainer currently defines `sameOrigin` as:

> a convenience boolean indicating whether the navigation is same-origin, and thus will stay in the same app history or not. (I.e., this is (`new URL(e.destination.url)).origin === self.origin`.)

However, there's a very closely related concept, which is "can you call `respondWith()`". This is not exactly the same. You can call `respondWith()` whenever the URL only differs in path/query/fragment components. (The same conditions as `history.pushState()`.)

The difference is cases like `https://example.com/foo` and `https://user:password@example.com/foo`. These are same-origin, but they differ in components beyond just path/query/fragment. So this means we cannot allow `navigate` handlers to intercept

```html
<a href="https://user:password@example.com/foo">click me</a>
```

on `https://example.com/` pages.

So: this means code such as

```js
if (event.sameOrigin) {
  event.respondWith(...)
}
```

will fail in such edge cases.

I see a few paths:

1. Keep the name and the semantics of `event.sameOrigin`. Code that is using credentialed URLs could get a failure when it calls `respondWith()`, but that's fine; it's a strange edge case. (Credentialed URLs are basically deprecated on the web.)

1. Keep the name of `event.sameOrigin`, but change the semantics to be about path/query/fragment-differences-only. This makes me uneasy because "same-origin" is a very well-defined concept and it'd be unfortunate to lie in such a way.

1. Change the name and semantics, to e.g. `event.canRespond`. This makes the code a bit less self-documenting, i.e. `if (event.canRespond) { event.respondWith(...) }` is less obvious to the reader than `if (event.sameOrigin) { event.respondWith(...) }`. But oh well.

1. Keep `event.sameOrigin` and also introduce `event.canRespond`.

Note that for (2) and (3), we lose the ability to easily know whether the destination will end up in the same app history list, since that _is_ origin-based. One would have to do the long test, i.e. `new URL(e.destination.url)).origin === self.origin`. I'm not sure if there's a strong use case for that though.

(Hmm, maybe we should change app history's scoping to use path/query/fragment-differences-only, instead of origin?? That gives the nice property that anything in your app history list, you can intercept with `respondWith()` and convert to a single-page nav... that change would be orthogonal though, I think.)

Anyway, I lean toward (3). Any other thoughts? /cc @annevk to see what he thinks of (2)...

### LINKED ISSUE #53: Forward/Back doesn't trigger navigate could cause restoration headaches
One primary use case for listening to events is to restore UI.

Imagine for instance, you push a new URL, and in a navigate handler respond to that URL change by rendering a new full page UI and  showing it on the page.

Then, say someone navigates back. How are you supposed to update your UI to remove the full page UI you added in response to the last URL update? Should you be listening to `currentchange`? If you do decide to listen to `currentchange` to handle this use case, how do you differentiate between back/forward navigations and other navigations that are handled by the `navigate` event handler?

I think there's a few options to solve this:
1. Do trigger `navigate` on back/forward but do NOT allow it to be intercepted.
2. Introduce a new event that handles this specific case (so it would be mutually exclusive with `navigate`), something like `navigatedirection`.
3. Add more fields to `currentchange` so you could know the cases where a `currentchange` is sent but a `navigate` is not .

There may be other options as well

domenic:
* Closes #53 by allowing respondWith() for same-document back/forward navigations (but not yet allowing preventDefault(); that's #32).
* Closes #51 by transitioning from event.sameOrigin to event.canRespond, which has updated semantics.
* Does not fire navigate events for document.open(), as per some initial implementation investigations that's more trouble than it's worth.
* Adds an example for special back/forward handling in the navigate event, per https://github.com/WICG/app-history/issues/53#issuecomment-789931864.
* Makes the appendix table a bit more precise about userInitiated, and expands it to cover cancelable and canRespond.

@natechapin and @tbondwilkinson, I'd love your review of these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,7 +369,7 @@ As for `document.open()`, it is a terrible legacy API with lots of strange side 
 
 Second, the following navigations **cannot be canceled** using `event.preventDefault()`, and as such will have `event.cancelable` equal to false:
 
-- User-initiated navigations (cross- or same-document) via the browser's back/forward buttons.
+- User-initiated same-document navigations via the browser's back/forward buttons.
 
 This is important to avoid abusive pages trapping the user by disabling their back button. Note that adding a same-origin restriction would not help here: imagine a user which navigates to `https://evil-in-disguise.example/`, and then clicks a link to `https://evil-in-disguise.example/2`. If `https://evil-in-disguise.example/2` were allowed to cancel same-origin browser back button navigations, they have effectively disabled the user's back button.
 
--- a/README.md
+++ b/README.md
@@ -313,10 +313,10 @@ Note how this example responds to various types of navigations:
 
 - Cross-origin navigations: let the browser handle it as usual.
 - Same-document fragment navigations: let the browser handle it as usual.
-- Same-document URL/state updates (via `history.pushState()`, `appHistory.update()`, etc.):
+- Same-document URL or state updates (via `history.pushState()` or `history.replaceState()`):
   1. Send the information about the URL/state update to `doSinglePageAppNav()`, which will use it to modify the current document.
   1. After that UI update is done, potentially asynchronously, notify the app and the browser about the navigation's success or failure.
-- Cross-document normal navigations:
+- Cross-document normal navigations (including those via `appHistory.push()` or `appHistory.update()`):
   1. Prevent the browser handling, which would unload the document and create a new one from the network.
   1. Instead, send the information about the navigation to `doSinglePageAppNav()`, which will use it to modify the current document.
   1. After that UI update is done, potentially asynchronously, notify the app and the browser about the navigation's success or failure.
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ First, the following navigations **will not fire `navigate`** at all:
 - User-initiated [cross-document](#appendix-types-of-navigations) navigations via browser UI, such as the URL bar, back/forward button, or bookmarks.
 - [`document.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/open), which can strip off the fragment from the current document's URL.
 
-User-initiated navigations of this sort are outside the scope of the webpage, and can never be intercepted or prevented, even if they are to same-origin documents. On the other hand, we do allow the page to intercept user-initiated _same_-document navigations via browser UI, e.g. if the user changes the fragment component in the URL bar.
+Navigations of the first sort are outside the scope of the webpage, and can never be intercepted or prevented. This is true even if they are to same-origin documents, e.g. if the browser is currently displaying `https://example.com/foo` and the user edits the URL bar to read `https://example.com/bar` and presses enter. On the other hand, we do allow the page to intercept user-initiated _same_-document navigations via browser UI, e.g. if the the browser is currently displaying `https://example.com/foo` and the user edits the URL bar to read `https://example.com/foo#fragment` and presses enter.
 
 As for `document.open()`, it is a terrible legacy API with lots of strange side effects, which makes supporting it not worth the implementation cost. Modern sites which use the app history API should never be using `document.open()`.
 
--- a/README.md
+++ b/README.md
@@ -1086,8 +1086,8 @@ Here's a summary table:
 |Browser UI (non-back/forward other)|Always cross|No|—|—|—|
 |`<a>`/`<area>`|Either|Yes|Yes ‡|Yes|Yes *|
 |`<form>`|Either|Yes|Yes ‡|Yes|Yes *|
-|`<meta http-equiv="refresh">`|Either|Yes|No|Yes|Yes *|
-|`Refresh` header|Either|Yes|No|Yes|Yes *|
+|`<meta http-equiv="refresh">`|Either ◊|Yes|No|Yes|Yes *|
+|`Refresh` header|Either ◊|Yes|No|Yes|Yes *|
 |`window.location`|Either|Yes|No|Yes|Yes *|
 |`window.open(url, name)`|Either|Yes|No|Yes|Yes *|
 |`history.{back,forward,go}()`|Either|Yes|No|Yes|Yes †*|
@@ -1099,6 +1099,7 @@ Here's a summary table:
 - † = No if cross-document
 - ‡ = No if triggered via, e.g., `element.click()`
 - \* = No if the URL differs in components besides path/fragment/query
+- ◊ = fragment navigations initiated by `<meta http-equiv="refresh">` or the `Refresh` header are only same-document in some browsers: [whatwg/html#6451](https://github.com/whatwg/html/issues/6451)
 
 See the discussion on [restrictions](#restrictions-on-firing-canceling-and-responding) to understand the reasons why the last few columns are filled out in the way they are.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -295,8 +282,8 @@ The following is the kind of code you might see in an application or framework's
 
 ```js
-  // Don't intercept cross-origin navigations; let the browser handle those normally.
-  if (!e.sameOrigin) {
+  // Some navigations, e.g. cross-origin navigations, we cannot intercept. Let the browser handle those normally.
+  if (!e.canRespond) {
```
tbondwilkinson, line 286: I had to think for a while before remembering that you can still cancel these navigations, even if you can't respond to them. Maybe an example? Otherwise some people may wonder why this is even needed

README.md
```diff
@@ -340,6 +327,61 @@ Note how this example responds to various types of navigations:
...
+  if (!e.canRespond || e.hashChange) {
+    return;
+  }
+
+  if (myFramework.currentPage) {
+    await myFramework.currentPage.transitionOut();
+  }
+
+  const isBackForward = appHistory.entries.includes(event.destination);
+  let { key } = event.destination;
```
tbondwilkinson, line 346: hypernit, `let {key}` is more common in the JS code I see, but maybe this is more common in specs?

README.md
```diff
@@ -340,6 +327,61 @@ Note how this example responds to various types of navigations:
...
+#### Restrictions on firing, canceling, and responding
+
+There are many types of navigations a given page can experience; see [this appendix](#appendix-types-of-navigations) for a full breakdown. Some of these need to be treated specially for the purposes of the navigate event.
+
+First, the following navigations **will not fire `navigate`** at all:
+
+- User-initiated [cross-document](#appendix-types-of-navigations) navigations via browser UI, such as the URL bar, back/forward button, or bookmarks.
+- [`document.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/open), which can strip off the fragment from the current document's URL.
+
+User-initiated navigations of this sort are outside the scope of the webpage, and can never be intercepted or prevented, even if they are to same-origin documents. On the other hand, we do allow the page to intercept user-initiated _same_-document navigations via browser UI, e.g. if the user changes the fragment component in the URL bar.
```
tbondwilkinson, line 366: This "even if they are to same-origin documents" only refers to `document.open()` right? If so, move it into that bullet, otherwise it's confusing because it looks like it's also possible in bullet 1

tbondwilkinson, line 366: What about clicking a cross-document link?

domenic: No, this is meant to cover cases like:

1. You are on `https://example.com/foo`
2. The user types `https://example.com/bar` in the URL bar

tbondwilkinson: ah okay I got confused by same-origin same-document again :)

domenic: Yeah, that's fine and interceptable. I guess I messed up by abbreviating "User-initiated cross-document navigations via browser UI" to "User-initiated navigations of this sort". This is meant to be only about the first bullet.

README.md
domenic: https://github.com/WICG/app-history#example-single-page-app-redirects is kind of that example. I split the examples so the most central ones were first and the less central ones were after the explanatory subsections, but maybe that was mistake.

README.md
```diff
@@ -340,6 +327,61 @@ Note how this example responds to various types of navigations:
...
+});
+```
+
+#### Restrictions on firing, canceling, and responding
+
+There are many types of navigations a given page can experience; see [this appendix](#appendix-types-of-navigations) for a full breakdown. Some of these need to be treated specially for the purposes of the navigate event.
+
+First, the following navigations **will not fire `navigate`** at all:
+
+- User-initiated [cross-document](#appendix-types-of-navigations) navigations via browser UI, such as the URL bar, back/forward button, or bookmarks.
```
frehner, line 363: I may be reading this wrong, but you list `user-initiated cross-document navigations via... back/forward button` as something that will never fire `navigate` here. 

yet below on line 372 it says `user-initiated navigations (cross- or same-document) via browser's back/foward buttons` which implies that navigate is fired (else there would be no need to mention `event.preventDefault()` ?)

also the table appears to list browser back/forward as firing navigate.

frehner (commented):
small clarification

README.md
domenic: Good catch. Listing cross-document browser back/forward navigations as non-cancelable is redundant since they won't fire navigate at all. Fixed.

As for the table, it says "Yes †*" where - † = No if cross-document (and \* = No if the URL differs in components besides path/fragment/query). I struggled with these "Yes" table entries; most of them are "Yes except in specific rarer circumstances", but that cell has more serious exceptions. Any suggestions are welcome.

README.md
```diff
@@ -1035,28 +1077,30 @@ Most navigations are cross-document navigations. Same-document navigations can h
...
-
-- Today it is not possible to intercept cases where other frames or windows programatically navigate your frame, e.g. via `window.open(url, name)`, or `history.back()` happening in a subframe. So, firing the `navigate` event and allowing interception in such cases represents a new capability. We believe this is OK, but will report back after some implementation experience. See also [#32](https://github.com/WICG/app-history/issues/32).
+|Trigger|Cross- vs. same-document|Fires `navigate`?|`e.userInitiated`|`e.cancelable`|`e.canRespond`|
+|-------|------------------------|-----------------|-----------------|--------------|--------------|
+|Browser UI (back/forward)|Either|Yes|Yes|No|Yes †*|
+|Browser UI (non-back/forward fragment change only)|Always same|Yes|Yes|Yes|Yes *|
+|Browser UI (non-back/forward other)|Always cross|No|—|—|—|
+|`<a>`/`<area>`|Either|Yes|Yes ‡|Yes|Yes *|
+|`<form>`|Either|Yes|Yes ‡|Yes|Yes *|
+|`<meta http-equiv="refresh">`|Either|Yes|No|Yes|Yes *|
```
natechapin, line 1087: Drive-by nitpick: meta Refresh and the Refresh header will always be cross-document, since they're effectively a reload. Same with form submission.

natechapin (approved):

README.md
```diff
@@ -340,6 +327,61 @@ Note how this example responds to various types of navigations:
 
 Notice also how by passing through the `AbortSignal` found in `e.signal`, we ensure that any aborted navigations abort the associated fetch as well.
 
+#### Example: async transitions with special back/forward handling
+
+Sometimes it's desirable to handle back/forward navigations specially, e.g. reusing cached views by transitioning them onto the screen. This can be done by branching as follows:
+
+```js
+appHistory.addEventListener("navigate", e => {
```
frehner, line 335: I think you need async here?

```suggestion
appHistory.addEventListener("navigate", async (e) => {
```

domenic: Oh, the problem here is I need to wrap this all in respondWith.

README.md
domenic: Not true in Firefox, it turns out! https://boom-bath.glitch.me/meta-refresh.html fun times.

Issue #13, opened 2021-02-03
Consider cancelling pending forward navigations when going back


esprehn:
Queued forward navigations should automatically cancel when executing a backwards navigation.

ex. In the photo gallery example:

1. Start on /photos/2
2. Click Next button which starts a navigation to /photos/3 with pushNewEntry(callback)
2. Navigation event fires and respondWith() starts a 500ms animation
3. Click Next 9 more times queuing 9 callbacks with pushNewEntry(callback)
4. Click Back button taking you to /photos/1
5. No more navigations happen, all 9 callbacks are cancelled.

The expected user behavior is not that the app goes back to /photos/1 and then jumps forward 9 times, but that all pending forward navigations are cancelled whenever going backwards.

I don't think this should be done manually through an AbortSignal. In practice the queued navigations are likely to come out of third party libraries or unrelated components.

tbondwilkinson:
Could this be handled by the router itself? Or are you suggesting this is universally desirable? 

I think the router (which knows about pending navigations) should be responsible for cancelling the push navigations on a back click. I'm not convinced the framework itself should do this behavior.

esprehn:
@tbondwilkinson I think this does need to be done by the platform because navigations come from multiple uncoordinated libraries on the page. 

I'm not sure I understand what you mean by `the router` knowing about pending navigations. There's no router here, just calls to pushNewEntry() from libraries all over the page. The spec does not expose a way to cancel a navigation, or get the list of pending navigations currently. You could imagine adding that of course, though I think this behavior makes sense as the default because I can't imagine the common scenario is one where users expect queued forward navigations to continue after clicking back. No existing web apps or native apps I'm aware of do that?

tbondwilkinson:
I do think that most applications will choose to have one central `navigate` listener installed on the page that will do the basic work of routing, even if there are multiple uncoordinated libraries on the page.

esprehn:
@tbondwilkinson I think you're misunderstanding the ticket here. The spec provides no facility for that central navigate listener to inspect the pending navigation stack or to cancel a pending navigation. Calls to pushNewEntry() go into a queue you can't interact with. There's also no way to know when the sequence of navigate events comes in should you want to cancel them, and that starts to make the router abstraction fairly complex.

I think requiring a sophisticated router that is observing all forward and backward navigation and picking and choosing what to cancel would be unfortunate and likely leads to a lot of buggy web apps compared to native. I don't think any real app jumps forward like this after going back.

[...]

Issue #11, opened 2021-02-03
How to detect or await a pending navigation if you didn't initiate it?


esprehn:
ex. Gmail uses a global queue, if there's a pending navigation it won't process any future requests to navigate until that's done. The opposite of the photo gallery example. If you click Next Page 20 times, nothing happens until the next page has loaded.

tbondwilkinson:
I do think that this type of coordination code should be done by the application itself- which assumedly has a navigation handler itself, and has probably some type of router that's installed via that route.

So, the router knows about all pending navigations, and if some part of the application needs to also know, it should ask the router.

Figuring out how this works on pages where some piece of logic CANNOT talk to the central router feels out of scope, and is a case we've punted on in other categories as well.

esprehn:
@tbondwilkinson I don't think it's reasonable to require a central router abstraction over top of the platform in this case. That's a major issue with the existing API surface of pushState() where ex. code that doesn't use react-router will break the app.

I don't think it's reasonable to say that no part of the page can call pushNewEntry() directly, and instead app components must go through some router library. That's also antithetical to the concept of web components where you can have two unrelated frameworks on the page like React and Lit and your app will still work.

tbondwilkinson:
Sorry, let me be more clear.

I think that anyone can call `pushNewEntry()` directly, but there needs to be some central logic that handles the conducting of navigations (deciding whether to cancel, etc.) and that would be the "router" that's installed via the navigate event listener. I don't think you'd want to load some third-party code on your page that could then reject navigations from other parts of the page, and especially the "hosting" page.

I do agree there's still some work to be done to enable third-party code to work with history, which someone brought up in #18. I don't think that any app component needs to be aware of what the centralized router implementation is, and should be able to use the native framework itself to get all information and make navigation requests as necessary. But I also don't think that any part of the app component needs to be listening directly to ALL navigations when they only really care about a portion of the URL or a portion of the history state.

domenic:
https://github.com/WICG/app-history/pull/68 attempts to address many of the surrounding issues around queued-up navigations, by getting rid of them: new navigations interrupt ongoing ones.

However, the essential problem this issue gets at in its OP still remains, which is: what if I *want* to wait on an ongoing navigation? How do I do that?

You can do so in the post-#68 explainer by listening for the `navigatefinish` + `navigateerror` events, and only doing work after those. However, this is not very ergonomic. I think a promise would be much nicer: something like

```js
[...]

Issue #10, opened 2021-02-03
Unclear use cases and example for photo gallery navigation queuing


esprehn:
https://github.com/WICG/app-history/blob/main/README.md#queued-up-single-page-navigations

Perhaps this is just unclear in the explainer, but if each click does pushNewEntry, and future ones don't run until the previous is finished with respondWith(), then does that mean there's a queue of [ push(1), push(2), push(3), ... push(N) ]? So if there's a 100ms animation, and I click next 20 times, it'll take 2 seconds to reach the right photo?

In a real application I think we'd probably want to skip ahead if you clicked Next many times or decrease the animation duration, instead of running all the animations for each one, but there's no way to know that the queue has become super deep to skip head from what I see in the API? Is it expected that consumers will build their own queue on the side to keep track of pending navigations?

It also took me a second to realize that the expected behavior is that the "Next" link has an anchor that points to the wrong photo number during a queued transition. So if you tap Next 10 times, and then on the 11th Shift+Click Next you'll open photo 2 in a new tab, not photo 11. I suppose maybe that is expected behavior, but that's kind of subtle. Is there precedent for either behavior on the web or native apps?

tbondwilkinson:
I think that applications should definitely do debouncing on the UI element itself.

If clicking some UI element triggers an expensive operation (like an animation, or a server fetch), you shouldn't allow clicking on it until you're sure that a previous expensive operation.

To make the navigation framework handle what are essentially bad inputs seems to me not the business of the framework

esprehn:
I don't think these are bad inputs. It's very common to allow someone to click Next rapidly in a photo gallery.

tbondwilkinson:
I think this is also related your other issue #13 - the broader question being "how much should the browser framework do automatically with discarding queued navigations"?

But about this specific use case, I think it's pretty tough to know from a framework perspective how to deal with queued navigations, and it's definitely an application decision whether new navigations should preempt existing ones, whether they should be deferred until after the current one finishes, or whether they should be thrown away altogether. I think that choice really depends on what the currently running navigation is.

PR #60, opened 2021-03-05
Do not roll back rejected-promise navigations

### LINKED ISSUE #47: Handling failed single-page app navs
The design of [navigation interception](https://github.com/WICG/app-history#navigation-monitoring-and-interception), especially after #46 lands, leads to a novel situation. Namely, you can perform a single-page app navigation which fails, by passing a rejected promise to `event.respondWith()`. A somewhat-realistic example might be:

```js
  e.respondWith((async () => {
    const response = await fetch("https://nonexistant.invalid/some-data.json");
    // ... etc ...
  })());
});

// This will get intercepted and converted to a failure.
location.href = "/foo";
```

The current plan as of #46 is that, because of the issues discussed in #19, all such navigations will synchronously succeed, but then asynchronously fail. Roughly:

1. `location.href`, the URL bar, etc. immediately start pointing to `/foo`, as if you'd done `history.pushState(null, null, "/foo")`.
2. `appHistory.current`, as well as the underlying session history model, gets pointed to a new history entry for `/foo`.
3. Any forward history entries get thrown away.
4. But then, once the browser sees that the promise is rejected, it "rolls back" the navigation:
    1. It sets `location.href`, the URL bar, etc. back to the original URL that initiated the navigation.
    1. `appHistory.current`, as well as the previous session history model, navigates back to the previous session history entry.
    1. The history entry for `/foo` created in step (2) gets thrown away.
    1. The forward history entries thrown away in (3) remain dead.

**Is this the right model for when the `navigate` handler tells us the navigation failed?**

The main alternative, I think, is to do a lot less. I.e., do not try to provide a rollback; leave `location.href`, the URL bar, etc. on `/foo`. Instead just signal to the app (e.g. with an event on `window.appHistory` and/or `window.appHistory.current`) that the promise rejected, and have the app handle that. The app could do its own rollback, perhaps. Or it could display a literal error page, like it might for a server-side 404.

I'm kind of leaning toward this do-less model. What do people think?

The other alternative would be to go the route of (2) in https://github.com/WICG/app-history/issues/19#issuecomment-785499230, which would allow us to delay all the updating stuff until we're sure the promise fulfills. Then there would be no navigate-then-rollback; we'd just never move at all in a failure case.

domenic:
This removes some of the complexity introduced in 9d750dbe7e061bbbcf98e6b6eee179e9ee46ac4a. Closes #47.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ appHistory.addEventListener("currentchange", e => {
     - [Restrictions on firing, canceling, and responding](#restrictions-on-firing-canceling-and-responding)
     - [Accessibility benefits of standardized single-page navigations](#accessibility-benefits-of-standardized-single-page-navigations)
     - [Measuring standardized single-page navigations](#measuring-standardized-single-page-navigations)
+    - [Example: handling failed navigations](#example-handling-failed-navigations)
     - [Example: single-page app "redirects"](#example-single-page-app-redirects)
     - [Example: cross-origin affiliate links](#example-cross-origin-affiliate-links)
   - [New navigation APIs](#new-navigation-apis)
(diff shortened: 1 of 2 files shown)


frehner:
Ah, I think this mostly answers a question I've had:

Given the following situation with multiple `navigate` event handlers:

```js
appHistory.addEventListener('navigate', e => {
  e.respondWith(Promise.reject())
})

appHistory.addEventListener('navigate', e => {
  e.respondWith((async () => {
    doPotentiallyExpensiveWork()
    // for API calls there is `e.signal`, so that case is covered
  }))
})
```

I had been wondering if the second event handler could somehow get notified that the navigation was cancelled and avoid calling `doPotentiallyExpensiveWork()`

But if I understand this change correctly, a rejected promise in `respondWith` will still navigate to the url; so `doPotentiallyExpensiveWork()` is still useful there (as opposed to how the spec, before this PR, was to "roll back" the navigation and render the work done in `doPotentiallyExpensiveWork()` useless. 

Does that sound correct?

domenic:
Yeah, that's another good point in favor of this less-complicated design! Although I'll admit that generally dealing with multiple concurrent or queued-up navigations is still a bit up in the air; we have some idea that signal will cancel ongoing ones, but the issues in #10, #11, and #13 are not yet resolved.

PR #61, opened 2021-03-05
Remove setState() and statechange

### LINKED ISSUE #57: Should we allow setState() separate from update()?
Continuing from @tbondwilkinson's comment at https://github.com/WICG/app-history/pull/54#issuecomment-789905258:

> I'm not sure about `setState()`.
>
> I think it creates the same problem that we have today with `history.pushState()` where it's updating the browser about the current state of things AFTER that state has been changed. So on the one hand we're encouraging people to do navigations by storing information in history and responding to those data changes in the navigate event, but this new `setState()` method is an out - it basically means that there's a workaround to update things in history without ever triggering a navigate event (where we would expect most people's rendering code to be). I think that dilutes the model and is pretty niche.

I'm sympathetic to these concerns. I think the crucial question is whether we consider a no-URL-update change to the state to be a navigation, or not.

In particular, the discussions in https://github.com/WICG/app-history/issues/42#issuecomment-786117352 indicated to me that maybe state-but-not-URL updates are conceptually different. E.g., they would be used for syncing state from the DOM to the `state` object, and it'd be strange for them to be considered navigations. The design in the explainer after #54 makes it so that `update()` always causes a full navigation by default, which the app would intercept with `navigate`. It then expects you to use `setState()` for other cases.

I may have jumped the gun here though in assuming they are different. If we think that an app will generally want to treat no-URL-update state changes as navigations on the same level as URL-update ones, then we should get rid of `setState()` and force everything through `update()`.

> Further, it means that some part of the application can just clear your state and you don't get any event about it besides `currentchange`. That also feels bad.

A late-breaking addition to #54 was the `statechange` event, so you get that. And to be clear, you *don't* get `currentchange` as formulated. Whether `currentchange` fires or not is tied to a few questions:

- What would be most useful? I still am not sure exactly when people will use `currentchange`, since `navigate` is so powerful and per-entry events cover other cases. See  #14.
- Do we semantically consider `currentchange` to mean "the current entry changed to a different entry than it was before" or "something about the current entry changed"? This is very related to #7.

domenic:
Partially reverts some of the changes in 3eee614dd6fcfa0a4f8620e2d2ea6a5799a5d37a, which upon further consideration were deemed not so good. Closes #57.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,10 +186,10 @@ The entry point for the app history API is `window.appHistory`. Let's start with
 
 - `sameDocument`: a boolean indicating whether this entry is for the current document, or whether navigating to it will require a full navigation (either from the network, or from the browser's back/forward cache). Note: for `appHistory.current`, this will always be `true`.
 
-It also has two methods, `getState()` and `setState()`, which retrieve the app history state for the entry. This is somewhat similar to `history.state`, but it will survive fragment navigations. Note that `getState()` always returns a fresh clone of the state:
+It also has a method `getState()`, which retrieve the app history state for the entry. This is somewhat similar to `history.state`, but it will survive fragment navigations, and `getState()` always returns a fresh clone of the state to avoid the [misleading nature of `history.state`](https://github.com/WICG/app-history/issues/36):
 
 ```js
-appHistory.current.setState({ test: 2 });
+appHistory.update({ state: { test: 2 } });
 
 // Don't do this: it won't be saved to the stored state.
 appHistory.current.getState().test = 3;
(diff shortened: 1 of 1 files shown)


frehner:
For clarity: with this PR, does this mean that the only way to update an Entry's state is to make it the current entry (if it isn't already) and then call `appHistory.update({state: newState})`? Or can you do `appHistory.entries[index].state = newState`?

domenic:
Yeah, that's right. There is no `state` property anymore, just `getState()`.

PR #63, opened 2021-03-05
Initial draft of navigate event spec

domenic:
@natechapin this should hopefully help. (Check out the preview link that _should_ be auto-generated.) It's a bit more complex than your implementation because I want to preserve any pushState() or replaceState() arguments, and it tries to incorporate all the stuff that you're deferring for now.

<del>I don't yet have the monkeypatches that fire the event, so the "fire an event" algorithm might not be quite right. But I think I captured all the inputs. (There are a lot of them... maybe some can be collapsed, once we see the call sites.)</del>

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/63.html" title="Last updated on Mar 11, 2021, 7:09 PM UTC (1b9c6b2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/63/4cf5cd7...1b9c6b2.html" title="Last updated on Mar 11, 2021, 7:09 PM UTC (1b9c6b2)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -236,7 +236,7 @@ An {{AppHistoryNavigateEvent}} has an associated [=URL=] <dfn for="AppHistoryNav
 <hr>
 
 <div algorithm="fire a navigate event">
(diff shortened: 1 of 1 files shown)


domenic:
TODO after the weekend: see if we can collapse those six booleans into an enum, given that there are only 4 call sites.

domenic:
Note to self: I think this needs updates to handle https://github.com/WICG/app-history/pull/65

domenic:
OK, I think this is complete. @natechapin mind taking a look? You probably would most enjoy the preview link in the OP, instead of the raw diff.

domenic:
@natechapin found a bug while implementing: the "Modify the traverse the history by a delta algorithm" patch needs to pass hashChange in same-document cases where the hash differs.

spec.bs
```diff
@@ -16,6 +16,28 @@ Complain About: accidental-2119 yes, missing-example-ids yes
...
+<pre class="link-defaults">
+spec: html; type: element; text: a
+</pre>
+<pre class="anchors">
+spec: html; type: dfn; urlPrefix: https://html.spec.whatwg.org/multipage/
+  text: serialized state; url: history.html#serialized-state
+  for: URL and history update steps
+    text: serializedData; url: history.html#uhus-serializeddata
+    text: title; url: history.html#uhus-title
+    text: isPush; url: history.html#uhus-ispush
```
natechapin, line 31: Semi-related: It seems strange to me that the HTML spec both has _isPush_ and the "history handling behavior". The spec changes proposed here will greatly expand the use of _isPush_. Can/should we deprecate _isPush_ in favor of the "history handling behavior" as a cleanup step?

natechapin (commented):
This looks very reasonable to me. I'm sure I'll find some additional things to tweak as I continue to implement.

Just one cleanup suggestion/question for now.

spec.bs
domenic: Hmm, I can see both sides. History handling behavior is very tied to the full navigate lifecycle right now (navigate / all the history algorithms). Whereas the URL and history update steps are much smaller scoped. We could reuse history handling (and assert that it's only ever "default" or "replace")... and I guess that's closer to what Chrome does. Hmm.

PR #68, opened 2021-03-11
Re-do interrupted and aborted navigations

### LINKED ISSUE #10: Unclear use cases and example for photo gallery navigation queuing

### LINKED ISSUE #13: Consider cancelling pending forward navigations when going back

### LINKED ISSUE #38: Interaction between the stop button and navigate event
One of the biggest issues that I ran into with my attempts to implement asynchronous navigation on top of the existing history API is that there was not a browser-based method for cancelling an asynchronous navigation. For me, the ideal mechanism to do this would be for the browser to to swap the refresh button to a stop button like it does when navigating to another origin.

> It gives the browser clear insight into when a navigation is being handled as a single-page navigation, and the provided promise allows the browser to know how long the navigation takes, and whether or not it succeeds. We expect browsers to use these to update their own UI (including any loading indicators; see whatwg/fetch#19 and whatwg/html#330 for previous feature requests).

Does this expectation include showing a stop button that would reject the navigation promise or is it purely for showing a loading bar/spinner?

domenic:
Closes #38 by making it explicit that the stop button is meant to trigger the navigate event's AbortSignal.

Closes #10 by getting rid of queued-up navigations, their associated event, and their associated callback variants to update() and push(). Instead, navigations during other navigations interrupt the ongoing one. As the new example tries to show, this seems to work pretty well.

Closes #13 by canceling all pending navigations when another one starts, of which the case discussed there is one example.

This does not yet solve #11; I will comment over there momentarily.

/cc @esprehn; I'd love you to take a look if you have time, since this tries to address some of the issues you found with our current setup. Also @tbondwilkinson since it was him and I who originally brainstormed the queuing system that this largely discards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,10 +77,10 @@ appHistory.addEventListener("currentchange", e => {
     - [Example: handling failed navigations](#example-handling-failed-navigations)
     - [Example: single-page app "redirects"](#example-single-page-app-redirects)
     - [Example: cross-origin affiliate links](#example-cross-origin-affiliate-links)
+    - [Aborted navigations](#aborted-navigations)
   - [New navigation APIs](#new-navigation-apis)
     - [Example: using `navigateInfo`](#example-using-navigateinfo)
-    - [Navigations while a navigation is ongoing](#navigations-while-a-navigation-is-ongoing)
-    - [Queued up single-page navigations](#queued-up-single-page-navigations)
+    - [Example: next/previous buttons](#example-nextprevious-buttons)
   - [Per-entry events](#per-entry-events)
   - [Current entry change monitoring](#current-entry-change-monitoring)
   - [Complete event sequence](#complete-event-sequence)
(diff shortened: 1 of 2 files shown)


tbondwilkinson:
I think this might also obviate the need for a callback, though if you feel it's still ergonomic I think that's fine too.

domenic:
Yep, I removed it for now. If we come up with another use case for it then we can easily add it back.

domenic:
I'll go ahead and merge this, but if anyone has thoughts or reviews after the fact, feel free to comment!

README.md
```diff
@@ -776,9 +802,14 @@ Between the per-`AppHistoryEntry` events and the `window.appHistory` events, as
...
         1. `appHistory.current` fires `finish`.
-        1. `navigateerror` fires on `window.appHistory`.
+        1. `navigateerror` fires on `window.appHistory` with the rejection reason as its `error` property.
         1. Any loading spinner UI stops.
         1. If the process was initiated by a call to an `appHistory` API that returns a promise, then that promise gets rejected with the same rejection reason.
+    1. Alternately, if the navigation gets [aborted](#aborted-navigations) before either of those two things occur:
+        1. `appHistory.current.finished` stays `false`, and `appHistory.current` never fires the `finish` event.
+        1. `navigateerror` fires on `window.appHistory` with an `"AbortError"` `DOMException` as its `error` property.
+        1. Any loading spinner UI stops. (But potentially restarts, or maybe doesn't stop at all, if the navigation was aborted due to a second navigation starting.)
+        1. If the process was initiated by a call to an `appHistory` API that returns a promise, then that promise gets rejected with the same  with an `"AbortError"` `DOMException`.
```
frehner, line 812: >with the same  with an

bathos:
This is awesome. I was considering opening an issue about the queuing but hadn’t investigated enough of the ongoing convos yet to know if it would be redundant.

We discovered in our “bootleg AppHistory” (based mainly on what was here maybe a month ago) that we never seemed to want the original queuing behavior in practice. We ended up dropping it and switching to always-abort-whatever-is-ongoing — even adding "signal" on the NavigateEvent as here.

 I know it’s anecdotal, but different folks reaching similar API/design conclusions feels really positive.

Issue #78, opened 2021-03-16
Should we fire navigate for cross-document history.back()?


domenic:
(All of this applies to forward as well as back, but I'll only mention back for simplicity.)

The current [table](https://github.com/WICG/app-history#appendix-types-of-navigations) has slightly divergent entries for `history.back()` vs. browser-UI-initiated back navigations. Summarizing:

- Browser back, same-document: cancelable = false, canRespond = true
    - Rationale: cancelable = false prevents back-trapping. But see #32.
- Browser back, cross-document: no event
    - Rationale: cancelable = false per above, and canRespond = false because we can't let you convert cross-document history traversals into same-document ones. So, don't fire the event, since it's not actionable.
- history.back(), same-document: cancelable = true, canRespond = true
    - Rationale: unlike browser back, back-trapping prevention isn't important
- history.back(), cross-document: cancelable = true, canRespond = false
    - Rationale: unlike browser back, back-trapping prevention isn't important, so the event can fire and be canceled

Since this design, a couple of data points have come up:

- @natechapin reports that distinguishing `history.back()` from browser back is currently hard in Chrome, so it'd be kind of nice if we didn't have to.

- As part of the discussions in #75, we realized that `history.back()` might originate from another frame, and letting that be canceled could be weird. Although it doesn't have the security difficulty discussed there of exposing the navigated-to URL, it might be good to treat the situations similarly.

These might lean toward something more conservative like making `history.back()` behave the same as browser back: never cancelable, and no event at all when it's cross-document:

- Browser back, same-document: cancelable = false, canRespond = true
- Browser back, cross-document: no event
- history.back(), same-document: cancelable = false, canRespond = true
- history.back(), cross-document: no event

However, #32's use case of allowing pages to show "are you sure you want to leave?", and in particular some of the latest ideas in https://github.com/WICG/app-history/issues/32#issuecomment-789944257, tugs us in a different direction, even if we did want to unify: something like

- Browser back, same-document: cancelable = true once per user activation, canRespond = true
- Browser back, cross-document: cancelable = true once per user activation, canRespond = false
- history.back(), same-document: cancelable = true once per user activation, canRespond = true
- history.back(), cross-document: cancelable = true once per user activation, canRespond = false

And at that point tying `history.back()` to user activation seems pretty strange, so I'd suggest

- Browser back, same-document: cancelable = true once per user activation, canRespond = true
- Browser back, cross-document: cancelable = true once per user activation, canRespond = false
[...]

PR #80, opened 2021-03-18
Restrict navigate firing when another window initiates the navigation

### LINKED ISSUE #75: Probably don't allow interception of navigations from other documents
Chrome implementation folks are concerned about the security implications of exposing the navigated-to URL in cases like:

- `window.open(url, nameOfSomeWindow)`
- `<form target="nameOfSomeWindow">`
- `window['nameOfSomeWindow'].location.href = ...`

As noted in the explainer,

> Note that today it is not possible to intercept cases where other frames or windows programatically navigate your frame, e.g. via `window.open(url, name)`, or `history.back()` happening in a subframe. So, firing the navigate event and allowing interception in such cases represents a new capability. We believe this is OK, but will report back after some implementation experience.

The case for intercepting these was never super-strong, so we should probably start restricting them. It's not clear to me yet whether we should always restrict them, or only restrict them when cross-origin.

<del>Similarly, currently we fire a cancelable but non-respondWith-able `navigate` event for `history.back()` navigates which cause you to move cross-document. This is slightly hard to implement, and is inconsistent with how we don't fire such an event for browser UI back/forward. It was done because of my general tendency toward "fire `navigate` in as many cases as possible". But probably we should reconsider. Especially since I think it could get you into weird states if only one frame cancels a back navigation.</del> <ins>Moving to a new issue.</ins>

If you're a web developer and would be very sad about these sorts of restrictions on navigation interception, please let us know. I'm hopeful that these are edge cases that don't need `navigate` event support.

domenic:
Closes #75, although this leaves open potential issues with history traversal from other documents which we'll track in #78.

@natechapin to review. /cc @csreis as a FYI.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/80.html" title="Last updated on Mar 19, 2021, 1:55 PM UTC (4424052)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/80/172985f...4424052.html" title="Last updated on Mar 19, 2021, 1:55 PM UTC (4424052)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,9 +366,12 @@ There are many types of navigations a given page can experience; see [this appen
 First, the following navigations **will not fire `navigate`** at all:
 
 - User-initiated [cross-document](#appendix-types-of-navigations) navigations via browser UI, such as the URL bar, back/forward button, or bookmarks.
+- [Cross-document](#appendix-types-of-navigations) navigations initiated from other [cross origin-domain](https://html.spec.whatwg.org/multipage/origin.html#same-origin-domain) windows, e.g. via `window.open(url, nameOfYourWindow)`, or clicking on `<a href="..." target="nameOfYourWindow">`
 - [`document.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/open), which can strip off the fragment from the current document's URL.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1115,9 +1118,10 @@ The web platform has many ways of initiating a navigation. For the purposes of t
   - Bookmarks
 - `<a>` and `<area>` elements (both directly by users, and programmatically via `element.click()` etc.)
 - `<form>` elements (both directly by users, and programmatically via `element.submit()` etc.)
+- As as special case of the above, the `target="nameOfSomeWindow"` attribute on `<a>`, `<area>`, and `<form>` will navigate a window whose `window.name` is `nameOfSomeWindow`
```
natechapin, line 1121: "as" -> "a"

natechapin (approved):

PR #84, opened 2021-03-19
Rename navigateTo → goTo and combine push/update → navigate

### LINKED ISSUE #81: Renaming navigateTo, push, and maybe update
An offline discussion with @annevk and @smaug---- indicated that `appHistory.navigateTo()` is a confusing name, as they expected it to take a URL (not a key). I tend to agree. @annevk suggested `go()` is a fine name, even though it takes a key instead of an offset like `history.go()` does. Maybe `goTo()` if we want to be clear it's something different.

Then, this would let us rename `push()` to `navigate()`, to better reflect its function. This is probably a good idea anyway after https://github.com/WICG/app-history/pull/54, so that people don't think that it behaves like `history.pushState()`.

Finally, they wondered if we could or should fold `update()` into this new `navigate()` function, e.g. `navigate("url", { replace: true })`. Again, after #54, which made `update()` more like `location.replace()`, this feels like like it might be a good idea.

domenic:
Closes #81.

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,7 +546,7 @@ Note how unlike `history.pushState()`, `appHistory.navigate()` will by default p
 
 Regardless of whether the navigation gets converted or not, calling `appHistory.navigate()` in this form will clear any future entries in the joint session history. (This includes entries coming from frame navigations, or cross-origin entries: so, it can have an impact beyond just the `appHistory.entries` list.)
 
-`appHistory.update` also takes a `replace` option, which indicates that it will replace the current history entry in a similar manner to `location.replace()`. It is used as follows:
+`appHistory.navigate()` also takes a `replace` option, which indicates that it will replace the current history entry in a similar manner to `location.replace()`. It is used as follows:
 
 ```js
 // Performs a navigation to the given URL, but replace the current history entry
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1223,14 +1223,10 @@ dictionary AppHistoryNavigationOptions {
   any navigateInfo;
 };
 
-dictionary AppHistoryPushOrUpdateOptions : AppHistoryNavigationOptions {
+dictionary AppHistoryNavigateOptions : AppHistoryNavigationOptions {
```
frehner, line 1226: should `boolean replace` be added to this dictionary?

README.md
```diff
@@ -1187,13 +1190,10 @@ interface AppHistory : EventTarget {
   readonly attribute boolean canGoBack;
   readonly attribute boolean canGoForward;
 
-  Promise<undefined> update(USVString url, optional AppHistoryPushOrUpdateOptions options = {});
-  Promise<undefined> update(optional AppHistoryPushOrUpdateFullOptions options = {}); // one member required: see issue #52
-
-  Promise<undefined> push(USVString url, optional AppHistoryPushOrUpdateOptions options = {});
-  Promise<undefined> push(optional AppHistoryPushOrUpdateFullOptions options = {});
+  Promise<undefined> navigate(USVString url, optional AppHistoryNavigateOptions options = {});
+  Promise<undefined> navigate(optional AppHistoryNavigateOptions options = {}); // one member required: see issue #52
```
frehner, line 1194: Related to the comment here: it's probably invalid to just have `appHistory.navigate({replace:true})` right? In that case, one member was provided, but it's still invalid. 

Probably hard to put that into the WebIDL format though.

bathos: Not something I feel super strongly about, but my intuition would have been that `navigate({})` (or say `go(current.key)`) wouldn’t throw, but also wouldn’t do anything except return a fulfilled promise (or more specifically, one that fulfills at the earliest point at which it could be fulfilled if it had had an effect).

domenic: Interesting. I'll probably have it throw for now since it's easier to make it un-throw than not, but this is worth keeping in mind...

PR #87, opened 2021-03-30
Fire navigatesuccess/navigateerror even without respondWith()

domenic:
Follows the implementation in https://chromium-review.googlesource.com/c/chromium/src/+/2786205.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/87.html" title="Last updated on Mar 30, 2021, 8:51 PM UTC (b49dfa3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/87/44c51c1...b49dfa3.html" title="Last updated on Mar 30, 2021, 8:51 PM UTC (b49dfa3)">Diff</a>


frehner:
I'm not very good at reading the spec, and I also don't know if you want to account for this yet or not, but I thought I should ask - 

Does this change also factor in that, if the event's AbortSignal was aborted, that `navigatesuccess` also doesn't fire and that `navigateerror` gets an `AbortError`?

domenic:
Good question! However, the spec doesn't have any AbortSignal support yet :). That's coming soon!

natechapin (approved):

PR #88, opened 2021-03-30
Add appHistoryEntry.id and clarify object identity

### LINKED ISSUE #7: Update vs. replace, entry mutability, and keys
(Moved from https://github.com/slightlyoff/history_api/issues/21 by @StringEpsilon, summarized by me.)

### Update vs. replace

The current proposal has `appHistory.updateCurrentEntry({ state, url })`. Although it doesn't state so explicitly, the implication of this method name is that the same `AppHistoryEntry` will have its `state` and `url` values updated. That is, it makes `AppHistoryEntry`s mutable, at least when they are the current entry. Notably, the `key` value would stay the same. Call this the "mutable history entry model".

An alternative approach would be to have `appHistory.replaceCurrentEntry({ state, url })`. This would generate a new `AppHistoryEntry` with the provided `state` and `url`. It would probably have a new `key` (but see below). Call this the "immutable history entry model".

Also, note that the mutable model could have `appHistory.replaceCurrentEntry()` _as well as_ `appHistory.updateCurrentEntry()`. Probably that would just be confusing, but it makes conceptual sense.

(For the record: `window.history` takes a hybrid approach. The current `window.history` history entry's state is mutable (via `history.state`), but if you want to change the page's current URL, you need to replace the current history entry using `history.replaceState()`.)

In general in programming, immutability can make things easier with less to worry about. That might be the case here, so I'm leaning toward switching to an immutable history entry model.

### The meaning and uses of `key`

In @tbondwilkinson's work on creating nice application-level wrappers over `window.history`, they have (at least) two keys:

- A "unique identifier" for the history entry, which is changed whenever the entry is replaced (i.e. whenever the URL of the entry changes). (Does it change if the state changes??)
- A "slot key" for the history entry, which stays stable even when the entry is replaced.

His apps use the slot key to navigate around, using the equivalent of `appHistory.navigateTo(slotKey)`. Whereas the unique identifier is used to cache resources that are too big to store in `state`, or to otherwise correlate a particular history entry with some out of band resource.

The app history proposal, with `updateCurrentEntry()`, currently has only the "slot key". If we changed to the immutable history entry model, with `replaceCurrentEntry()`, we could either:

- keep `key` as the slot key (so, the new `AppHistoryEntry` would have the same `key` value as the old one), or
- switch `key` to be a unique identifier (so, the new `AppHistoryEntry` would generate a new `key` UUID).

We could also introduce both types of keys, or switch `key` to be a unique identifier and introduce `index`, or similar.

A potential issue with unique keys is that we need to define the behavior for what happens if you do something like

```js
const uniqueKey = appHistory.currentEntry.key;
appHistory.replaceCurrentEntry();
appHistory.navigateTo(uniqueKey);
```

i.e. if you try to navigate to a key which is no longer in the current history entries list. I think we'd just fail, probably in the same way `appHistory.back()` fails if you're at the beginning of the history list, or `appHistory.navigateTo("non-existant-key")` fails.

### Replacing and the old entry

If we did switch to an immutable history model with `appHistory.replaceCurrentEntry()`, we could consider adding a feature like `appHistory.currentEntry.replacedEntry` pointing to the now-replaced entry. This would extend indefinitely (?), so you could do `appHistory.currentEntry.replacedEntry.replacedEntry` etc.

This also helps with the potential data loss, where calling `replaceCurrentEntry()`, or `updateCurrentEntry()`, will destroy any `state` value you had there. On the other hand, maybe this sort of "data loss" is intentional, in that it's better not to use up memory and storage space on replaced entries. Hmm.

domenic:
Closes #7.

Would love a look from @tbondwilkinson in addition to @natechapin. With this in place I'll feel more comfortable speccing a basic `AppHistoryEntry` and `appHistory.current`.


frehner:
The section explaining the differences between key and id is great. 💯 

One thing that I just thought of, that probably isn't that important but I was curious about: does the AppHistoryEntry object reference stay the same in all instances? e.g. 

```js
const previous = appHistory.current
await appHistory.navigate({replace: true, state: ''})
const current = appHistory.current

console.log(previous.id === current.id) // false
console.log(previous.key === current.id) // true

console.log(previous === current) // ???
```

domenic:
The plan is for the identity to match `id`. (Except in cases where that's impossible and doesn't make sense, like browser session restore.)

The explainer tries to talk about this in "The complete list of ways the current app history entry can change to a new entry are:" but maybe it'd be best to be more explicit...

domenic:
TODO: talk about session restore in more detail. I think it should be fine to generate new IDs upon session restore (while keys stay the same).

tbondwilkinson:
Overall looks great.

I think I would be a little surprised if IDs are regenerated upon session restore, because there are lots of caches you might have stored history-correlated resources that are preserved across a session restore

Issue #59, opened 2021-03-05
currentchange is not suitable for measuring the time


domenic:
After #46, `appHistory.current` updates synchronously, and thus `currentchange` fires synchronously, for all single-page navigations. So the example in the explainer which uses `currentchange` to measure the time is totally wrong.

This should probably be solved together with #14 and #33 but I wanted to note it separately since it's a serious bug in the explainer.

tbondwilkinson:
A counter proposal I think would be a `afternavigate` (or `navigateend` or whatever) event, that is sent whenever a navigation is fully finished, that would have largely the same properties as the `currentchange` event but would have the added benefit of only firing after a navigation had fully settled. I think if we go this route, we would get rid of `currentchange` altogether, and changing the timing of the event I think actually makes it more broadly useful (since it's not just identical to navigate but without the respondWith capability).

frehner:
Was talking with someone and the confusion around the timing of `currentchange` and the differences to `navigate` came up, so I remembered this issue.

Regarding a new event - would it just be easier to just add the timing information to `navigatesuccess` - since that event fires when `respondWith()` promises complete (unless it's been aborted), right?

domenic:
Yeah, `navigatesuccess` has the appropriate timing for that. However per #33 we might want to expose the timing information in a completely different way; we'll see...

Issue #41, opened 2021-02-17
Retrieving data from replaced entries


domenic:
A nice feature of the navigation API is that (unless we implement #9) data is almost never "lost". This solves a lot of problems we hear about today where `history.state` is unreliable, or there's no way to easily know what your previous history entry was without keeping a side table, etc.

However, there is one place in the current design where data can be lost: if you use `navigation.navigate(url, { replace: true })`, `history.replaceState()`, or `location.replace()`, the current navigation API history entry data disappears.

I say "kind of" because object references get replaced, not updated. So you could have a reference to the old `NavigationHistoryEntry` object, which would allow you to access `url` and `getState()`. (But it's no longer in `appHistory.entries()`.)

We could pretty easily keep these replaced `NavigationHistoryEntry`s around and expose them. For example:

- Each app history entry could have a `replaced` property that is null, or the entry it replaced. (This allows potentially chaining, e.g. `navigation.currentEntry.replaced.replaced.replaced`.) This lets you access them forever. (Is this a problem for memory usage?)

- The `currententrychange` event could have a `replaced` property which contains the replaced `NavigationHistoryEntry`. This lets unrelated parts of the code (i.e., not the code doing `navigation.navigate(, { replace: true })`) grab important data as necessary.

tbondwilkinson:
The only other case where they got lost is if they're in the forward stack, or in most browsers if it's evicted by the back stack growing too long (I believe the limit is 40). I think the dispose event helps to find out when this happens though, so it's not a major concern.

Another possible solution is to identify in the dispose event that it was a replace, rather than an evict, which would allow you to grab a handle on the entry (and state) that's replacing yours.

yoavweiss:
> [!NOTE]
> 🤖 **This comment was written and posted by an AI agent** (Claude, running in the [Pi](https://github.com/badlogic/pi-mono) coding agent), acting on behalf of @yoavweiss while triaging this repository ahead of archiving it. Migration scope was agreed with @noamr. **If this is wrong, please say so and it can be reopened.**

**Closing as stale — not resolved.**

Recovering `state` and `url` from entries dropped by `replace` is still not possible. Noted as adjacent on the migrated umbrella issue whatwg/html#12811, since a general entry-manipulation design would need to decide what happens to replaced entries. If you want just the read side, that is worth its own focused issue. cc @domenic

[...]

Issue #33, opened 2021-02-11
What design should be used for reporting navigation duration?


domenic:
After #27, the proposal [discusses](https://github.com/WICG/app-history#current-entry-change-monitoring) how you can get the "duration of the navigation" via `event.timeStamp - event.startTime`, at least for same-document navigations. It has an open question about whether this should work for cross-document navigations (related to #31).

A larger question is whether this property-on-an-event design is the right approach at all. We need to coordinate with the WebPerf folks, e.g. @yoavweiss is apparently [working on](https://docs.google.com/presentation/d/1-gckm8YTJfDnXa_Td8_mWbFDEITtH5CsKOoRKwg4JIM/edit#slide=id.g9c56539c45_0_30) a way to generate `PerformanceEntry` instances for bfcache navigations, which have some similarities to same-document navigations. Maybe (probably?) we should just have same-document navigations be measured via the those performance timeline mechanisms as well.

kenchris:
The @w3ctag think that is makes sense to use the same primitive for this and bfcache navigations, but developers should be able to see what navigation type happened so that you can differentiate the two

Issue #14, opened 2021-02-03
Expanding examples and available information for currententrychange


domenic:
(Moved from https://github.com/slightlyoff/history_api/issues/24 by @SetTrend and https://github.com/slightlyoff/history_api/pull/30 by @matt-buland-sfdc; summarized and synthesized by me.)

The proposal's current [`currententrychange` section](https://github.com/WICG/app-history#current-entry-change-monitoring) only has one weak example, around measuring navigation time. But since it's meant as a replacement for some subset of what people use `popstate` for today, we should expand it with others. @matt-buland-sfdc suggests:

> - change the page/content being shown based on the newly accepted location
>  - especially for handling browser-back
> - show side-bar content relevant to the current page (e.g. help items in the current page, etc)

and @SetTrend suggests that examples like these could be done more easily if the event had a pointer to the previous entry; his code is

```js
appHistory.addEventListener("currententrychange", e => {
  let diff = e.currentEntry.state.listItems.length - e.previousEntry.state.listItems.length;

  document.getElementById("result").innerText = `${Math.Abs(diff)} items have been ${diff < 0 ? "removed" : "added"}`;
});
```

domenic:
I'm coming back to this and still struggling a bit. All of the cases described in the OP seem like they should be done in the `navigate` event; i.e., as part of intercepting the navigation and converting it to a single-page navigation, you should do your (possibly async) page-updating there. What role does that leave for `currentchange`?

It still feels like such an event should probably exist; changing something so central as `appHistory.current` seems like something you should be notified about! But I just can't find a use case. So I'll tag this as "feedback wanted", and if we can't find any use cases, maybe we should delete it.

matt-buland-sfdc:
The issue with updating the page in `navigate` is that hook doesn't get invoked upon historical navigations (e.g. browser back to a location in the SPA's domain). Unless you're suggesting folks implement `oncurrententrychange => navigate('.')`. Reacting to historical navigations via popstate has been a struggle for us, and `currententrychange` mostly resolves them, which can demonstrate a big benefit to others as well. My original ask was to identify "what are the best practices for maintaining and reacting to navigations to previous (historical) entries?"

domenic:
> is that hook doesn't get invoked upon historical navigations (e.g. browser back to a location in the SPA's domain).

It does.

matt-buland-sfdc:
[...]

PR #90, opened 2021-04-02
Add appHistory.transition with first-class rollback

### LINKED ISSUE #11: How to detect or await a pending navigation if you didn't initiate it?

### LINKED ISSUE #86: Better example and API for rolling back failed navigations
In an offline discussion @atscott pointed out that the rollback example at https://github.com/WICG/app-history#example-handling-failed-navigations does the wrong thing in any cases besides a simple non-replace forward navigation. Some cases to consider would be:

- When a back navigation results in a rejected promise and corresponding `navigateerror` event (e.g., after logging out, trying to go back to a login-guarded page, which the router rejects). Then, the correct way of rolling back the navigation is actually to do `appHistory.back()`
- When you skip multiple entries, e.g. if you go back two steps. Then the correct way of rolling back the navigation would involve some `appHistory.goTo()` with a saved key.
- When you perform a replace navigation that fails. Then the correct way of rolling back would probably be another replace, back to the old URL/state.

Also, he mentioned that such rollbacks likely want to communicate to the router that they are a rollback, so it can treat them specially. This is likely best done via a `navigateInfo` parameter.

There are a few ways one could envision solving this. So far I've come up with:

- Add an `event.previous` `AppHistoryEntry`, to the `navigateerror` event. (And to the `navigatesuccess` event?) Then rollbacks would be somewhat manual, e.g. `await appHistory.goTo(event.previous.key, { navigateInfo })`.
   - The failed-replace case would be especially tricky. I think you could detect it if we follow through on #7 (so that you'd have `event.previous.key === appHistory.current.key`, even though `event.previous.id !== appHistory.current.id`). But it'd need its own special-case coding, and it'd be easy to forget.
- Add a "do what I mean" `event.rollback({ navigateInfo })` to the `navigateerror` event.
- Do both of these, so that `event.rollback()` is explained as a helper function that does some of the trickier rollback logic for you, but you can always construct your own.
- Some variant of these that is not tied to the `navigateerror` event, e.g. we could just always have `appHistory.previous` available (which might also help with #41).

Any thoughts, or other ideas?

domenic:
This gives insight into any "ongoing" navigation, i.e. a navigation that hasn't yet reached navigationsuccess/navigationerror because the promise passed to respondWith() has not yet settled.

Additionally:

* Removes the finished property from every AppHistoryEntry, which is nice because it only ever really made sense on the current entry; it was about the transition. Instead, the presence or nullness of appHistory.transition can be used.
* Adds a type property to AppHistoryNavigateEvent since we're going to have it on appHistory.transition anyway so having it earlier (before respondWith() is called) seems very reasonable.

Closes #86. Closes #11 by adding the appHistory.transition.finished promise.

Helps with #41 as you can retrieve data from the replaced entry, at least during the transition, with appHistory.transition.previous.

Probably helps with #14 (although currentchange needs a bit of an overhaul) since appHistory.transition has the sort of useful information discussed there, such as the previous entry or the type of navigation.

---

I omitted the `initiated` timestamp for now because: (a) it's still unclear what we're doing with time measurement in general, per #33 and #59; (b) the "transition" was initiated when you called `respondWith()`, and that seems a bit less useful than the time that the user clicked their mouse or the time when you did `location.href` or whatever. Usually they'll be close, but especially with input lag not always...

Potential bikeshed points I'd be happy to get feedback on:

- `appHistory.transition` vs. `appHistory.ongoing` vs. `appHistory.currentNavigation` vs. ...
- `"traversal"` vs. `"go"`
- `appHistory.transition.previous` vs. `appHistory.transition.source` vs. ...


README.md
```diff
@@ -418,9 +422,47 @@ This isn't a complete panacea: in particular, such metrics are gameable by bad a
 
 - We hope that most analytics vendors will come to automatically track `navigate` events as page views, and measure their duration. Then, apps using such analytics vendors would have an incentive to keep their page view statistics meaningful, and thus be disincentivized to generate spurious navigations.
 
+#### Aborted navigations
+
+As shown in [the example above](#example-replacing-navigations-with-single-page-app-navigations), the `navigate` event come with an `event.signal` property that is an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal). This signal will transition to the aborted state if any of the following occur before the promise passed to `respondWith()` settles:
```
atscott, line 427:
```suggestion
As shown in [the example above](#example-replacing-navigations-with-single-page-app-navigations), the `navigate` event comes with an `event.signal` property that is an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal). This signal will transition to the aborted state if any of the following occur before the promise passed to `respondWith()` settles:
```

README.md
```diff
@@ -418,9 +422,47 @@ This isn't a complete panacea: in particular, such metrics are gameable by bad a
...
+
+- `type`: either `"replace"`, `"push"`, or `"traversal"` indicating what type of navigation this is
+- `previous`: the `AppHistoryEntry` that was the current one before the transition
+- `finished`: a promise which fulfills with undefined when the `navigatesuccess` event fires on `appHistory`, or rejects with the corresponding error when the `navigateerror` event fires on `appHistory`
+- `rollback()`: a promise-returning method which allows easy rollback to the `previous` entry
+
+Note that `appHistory.transition.rollback()` is not the same as `appHistory.back()`: for example, if the user goes back two steps, then `appHistory.rollback()` will actually go forward to steps. Similarly, it handles rolling back replace navigations by doing another replace back to the previous URL and app history state, and it rolls back push navigations by actually removing the entry that was previously pushed instead of leaving it there for the user to reach by pressing their forward button.
```
atscott, line 461:
```suggestion
Note that `appHistory.transition.rollback()` is not the same as `appHistory.back()`: for example, if the user goes back two steps, then `appHistory.rollback()` will actually go forward two steps. Similarly, it handles rolling back replace navigations by doing another replace back to the previous URL and app history state, and it rolls back push navigations by actually removing the entry that was previously pushed instead of leaving it there for the user to reach by pressing their forward button.
```

atscott, line 461: Does `rollback` trigger a `navigate` event? I think it would be good to mention that here either way.

tbondwilkinson, line 461: "if the user navigates two steps back"

"will actually go forward two steps"

tbondwilkinson, line 461: "it handles rolling back replace navigations by reverting back to the URL and app history state"

tbondwilkinson:
Yay bike-shedding:

1. I think all of these are pretty solid, though I still have a slight preference for "transition". I lean towards it because I like how the property access looks (`appHistory.transition.type`). If we want a combo, `ongoingTransition` or `currentTransition` is even more explicit.
2. `traversal` is a little wordier, but is probably more accurate. No strong preference. If we go with "traversal" I think it should be "traverse" since the other types are verby (push, replace, traverse).
3. Slight preference for "source" and "destination" terminology, though another option is "from" or "before". `appHistory.transition.from` is kinda nice.

README.md
domenic: Yep, it does! Will mention.

domenic:
"traverse" is definitely better than "traversal", so I've at least updated in that direction. I'm still willing to be swayed toward "go".

I like "from" a good bit better than "previous" so I'll switch to that for now.

frehner:
Pretty late to the party, but looking at implementing this and came across this question:

For replace navigations, e.g. `appHistory.navigate({replace: true, state: {}})`, what is AppHistory.transition.from? For example:

```js
AppHistory.transtion.from === AppHistory.current // true?
AppHistory.transtion.from.state === AppHistory.current.state // false?
```

etc.

frehner:
>Note that appHistory.transition.rollback() is not the same as appHistory.back(): for example, if the user navigates two steps back, then appHistory.rollback() will actually go forward two steps. Similarly, it handles rolling back replace navigations by reverting back to the previous URL and app history state. And it rolls back push navigations by actually removing the entry that was previously pushed, instead of leaving it there for the user to reach by pressing their forward button.
>
>Also note that appHistory.transition.rollback() will itself trigger a navigate event. This means you can continue to centralize your response to navigations in the navigate event handler, i.e. rollback-initiated pushes/replaces/traversals go down the same application code path as other pushes/replaces/traversals.

If you call `AppHistory.transition.rollback()`, it will fire a new `navigation` and thus a new `AppHistory.transition` will be created for that navigation event, correct?

If so, what is the `AppHistory.transition.type` that's associated with that navigation? It doesn't seem like `push | replace | transition` really encapsulates what is going on, since a rollback will do things that `push | replace | transition` will never do.

Should perhaps an additional `transition.type` be added specifically for `rollback`s?

frehner:
I like the idea of `AppHistory.transition.finished` being a promise, but as I've been working on it I think it can cause some undesired headaches as well. Specifically this part:

```
- If [the promise returned from navigate()] rejects, fire `navigateerror` on `appHistory` and reject `appHistory.transition.finished`.
```

If I'm understanding this correctly, it means that if someone wants to call `AppHistory.navigate()`, they are required to now have error handling for _both_ the promise returned from `navigate()` and `AppHistory.transition.finished`, right? Otherwise they will have unhandled promise rejections?

Additionally, since `AppHistory.transition` is cleared after each transition, you'll have to redo this error handling every time you call navigate.

For example:

```js
try {
  const navigationPromise = appHistory.navigate('/url')
  // since there are two promises we have to handle, and since .transition is not created until we call .navigate()
  // we have call .navigate() but not await it so we can then access .transition
  await Promise.all([navigationPromise, appHistory.transition.finished])
catch(err){
  // handle both, or one, or the other
}
```

Or am I misunderstanding this completely? I could be wrong here.

domenic:
> AppHistory.transtion.from === AppHistory.current // true?

Yeah, I think that's the idea!

> If you call AppHistory.transition.rollback(), it will fire a new navigation and thus a new AppHistory.transition will be created for that navigation event, correct?

Correct.

> If so, what is the AppHistory.transition.type that's associated with that navigation?

Rolling back a traverse (e.g. -2) is done with a traverse (e.g. +2). Rolling back a replace is done by replacing (with the previous URL/state/etc. values). And rolling back a push is done with a traverse (-1).

> I like the idea of AppHistory.transition.finished being a promise, but as I've been working on it I think it can cause some undesired headaches as well. Specifically this part:

Good catch. We'd mark the promise as handled (equivalent to immediately doing `appHistory.transition.finished.catch(() => {})` upon promise creation), as is done for the other cases like this on the web platform, notably in streams. So, you'd never get unhandled promise rejections from `appHistory.transition.finished`.

frehner:
> And rolling back a push is done with a traverse (-1).

With the caveat that the rolled-back entry is removed, and you can no longer go forward as well - so, slightly different behavior than a normal traversal. 

Not sure if that’s an important enough distinction for developers to know about or not though. 

> We'd mark the promise as handled (equivalent to immediately doing appHistory.transition.finished.catch(() => {}) upon promise creation)

Ah ok that sounds good. Thank you!

frehner:
Given what you said above

>Good catch. We'd mark the promise as handled (equivalent to immediately doing appHistory.transition.finished.catch(() => {}) upon promise creation), as is done for the other cases like this on the web platform, notably in streams. So, you'd never get unhandled promise rejections from appHistory.transition.finished.

Should the text in this PR be updated to remove parts where it says `appHistory.transition.finished` will reject? If so, I can create a new issue (or a PR to fix it?).

e.g. 

>If it rejects, fire `navigateerror` on `appHistory` and reject `appHistory.transition.finished`.


 >- `finished`: a promise which fulfills with undefined when the `navigatesuccess` event fires on `appHistory`, or rejects with the corresponding error when the `navigateerror` event fires on `appHistory`
 
 On a different note, I am curious what your thoughts are on my comment above:
 
 >And rolling back a push is done with a traverse (-1).
>>With the caveat that the rolled-back entry is removed, and you can no longer go forward as well - so, slightly different behavior than a normal (-1) traversal.
>>
>>Not sure if that’s an important enough distinction for developers to know about or not though.

PR #91, opened 2021-04-03
Specify appHistory.entries() and appHistory.current

domenic:
This includes most of the specification for AppHistoryEntry itself.

I haven't really done a self-review so there might be obvious holes. I am especially excited to compare this with https://chromium-review.googlesource.com/c/chromium/src/+/2803480 and see how they differ.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/91.html" title="Last updated on Apr 9, 2021, 7:27 PM UTC (91eef45)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/91/65701cd...91eef45.html" title="Last updated on Apr 9, 2021, 7:27 PM UTC (91eef45)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -23,25 +23,28 @@ Assume Explicit For: yes
 spec: html; type: element; text: a
 </pre>
 <pre class="anchors">
-spec: html; type: dfn; urlPrefix: https://html.spec.whatwg.org/multipage/
-  text: serialized state; url: history.html#serialized-state
-  text: session history; url: history.html#session-history
-  text: session history entry; url: history.html#session-history-entry
-  for: URL and history update steps
-    text: serializedData; url: history.html#uhus-serializeddata
-    text: isPush; url: history.html#uhus-ispush
-  for: session history
-    text: current entry; url: history.html#current-entry
-  for: session history entry
-    text: document; url: history.html#she-document
-    text: URL; url: history.html#she-url
-  for: history handling behavior
-    text: default; url: browsing-the-web.html#hh-default
-    text: replace; url: browsing-the-web.html#hh-replace
-    text: entry update; url: browsing-the-web.html#hh-entry-update
-  for: navigate
-    text: historyHandling; url: browsing-the-web.html#navigation-hh
-    text: navigationType; url: browsing-the-web.html#navigation-navigationtype
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
+  type: dfn
+    text: serialized state; url: history.html#serialized-state
+    text: session history; url: history.html#session-history
+    text: session history entry; url: history.html#session-history-entry
+    for: URL and history update steps
+      text: serializedData; url: history.html#uhus-serializeddata
+      text: isPush; url: history.html#uhus-ispush
+    for: session history
+      text: current entry; url: history.html#current-entry
+    for: session history entry
+      text: document; url: history.html#she-document
+      text: URL; url: history.html#she-url
+    for: history handling behavior
+      text: default; url: browsing-the-web.html#hh-default
+      text: replace; url: browsing-the-web.html#hh-replace
+      text: entry update; url: browsing-the-web.html#hh-entry-update
+    for: navigate
+      text: historyHandling; url: browsing-the-web.html#navigation-hh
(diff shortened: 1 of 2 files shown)


spec.bs
```diff
@@ -251,6 +267,8 @@ The following are the [=event handlers=] (and their corresponding [=event handle
   1. Set |appHistory|'s [=AppHistory/entry list=] to |newEntryList|.
 
   1. Set |appHistory|'s [=AppHistory/current index=] to |newCurrentIndex|.
+
+  <p class="note">For same-document navigations much of this algorithm becomes a no-op, as only the [=AppHistory/current index=] will ultimately be updated. Implementations can special-case same-document navigations and avoid reassembling the [=AppHistory/entry list=].
```
domenic, line 271: This is not true because you need to update the index member.

spec.bs
```diff
@@ -427,3 +706,49 @@ With the above infrastructure in place, we can actually fire and handle the {{Ap
     1. Let |continue| be the result of [=firing a navigate event=] at |appHistory| given <var ignore>url</var>, with <i>[=fire a navigate event/isPush=]</i> set to false, <i>[=fire a navigate event/isSameDocument=]</i> set to |isSameDocument|, <i>[=fire a navigate event/userInvolvement=]</i> set to |userInvolvement|, and <i>[=fire a navigate event/isHistoryTraversal=]</i> set to true.
     1. If |continue| is false, abort these steps.
 </div>
+
+<h2 id="session-history-patches">Patches to session history</h2>
+
+This section details monkeyaptches to [[!HTML]] to track appropriate data for associating an {{AppHistory}} with a [=session history entry=].
```
natechapin, line 712: monkeyaptches -> monkeypatches

spec.bs
```diff
@@ -427,3 +706,49 @@ With the above infrastructure in place, we can actually fire and handle the {{Ap
...
+
+* <dfn for="session history entry">app history state</dfn>, which is [=serialized state=] or null, initially null
+
+<h3 id="session-history-patches-key">Carrying over the app history key</h3>
+
+Update the <a spec="HTML">update the session history with the new page</a> algorithm's "<a for="history handling behavior">`replace`</a>" case to set <var ignore>newEntry</var>'s [=session history entry/app history key=] to <var ignore>sessionHistory</var>'s [=session history/current entry=]'s [=session history entry/app history key=].
+
+<h3 id="session-history-patches-origin">Tracking the [=session history entry/origin=] member</h3>
+
+Update the <a spec="HTML">update the session history with the new page</a> algorithm to set the [=session history entry=]'s [=session history entry/origin=] to <var ignore>newDocument</var>'s [=Document/origin=].
```
natechapin, line 732: This should mention that we only copy the key if origins match?

spec.bs
```diff
@@ -139,6 +229,85 @@ The following are the [=event handlers=] (and their corresponding [=event handle
...
+
+    1. Let |she| be |sessionHistory|[|forwardIndex|].
+
+    1. If |she|'s [=session history entry/origin=] is [=same origin=] with |currentSHE|'s [=session history entry/origin=], then [=list/append=] |she| to |appHistorySHEs|.
+
+    1. Otherwise, [=iteration/break=].
+
+    1. Set |forwardIndex| to |forwardIndex| + 1.
+
+  1. Let |newCurrentIndex| be the index of |currentSHE| within |appHistorySHEs|.
```
natechapin, line 278: I don't think `currentShe` has been inserted into `appHistorySHEs` by this point?

domenic: It was appended during the forward loop, since we set forwardIndex initially to the index of currentSHE within sessionHistoryEntry. Is that too subtle? Maybe there should be a dedicated step for inserting currentSHE?

spec.bs
domenic: Great catch; done, thanks.

spec.bs
natechapin: I think either a separate state or a non-normative note might be helpful. Or maybe I'm just biased by my implementation :)

domenic: Done. It is indeed nice and symmetrical now.

PR #93, opened 2021-04-08
Spec navigationType and fix intercepted history traversal spec

domenic:
This renames the explainer's "type" property to "navigationType", since event.type already exists. It then specifies how to populate it.

More importantly, this fixes an issue where calling respondWith() would convert any history traversal into a replace navigation, which was really broken.

---

I'm considering whether to refactor "fire a navigate event" into two algorithms, one for history traversal and one for otherwise, since right now it has a bunch of optional parameters some of which only make sense for one or the other path.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/93.html" title="Last updated on Apr 9, 2021, 5:06 PM UTC (aa9cef1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/93/a48cf5a...aa9cef1.html" title="Last updated on Apr 9, 2021, 5:06 PM UTC (aa9cef1)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ The event object has several useful properties:
 
 - `canRespond`: indicates whether `respondWith()`, discussed below, is allowed for this navigation.
 
-- `type`: either `"push"`, `"replace"`, or `"traverse"`.
+- `navigationType`: either `"push"`, `"replace"`, or `"traverse"`.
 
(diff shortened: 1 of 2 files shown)


domenic:
> I'm considering whether to refactor "fire a navigate event" into two algorithms, one for history traversal and one for otherwise, since right now it has a bunch of optional parameters some of which only make sense for one or the other path.

Upon thinking a bit more, I'd like to delay that until we start populating `event.destination`. That will probably impact these call sites a good deal as well.

natechapin (approved):

Issue #95, opened 2021-04-14
Should appHistory.navigate()'s returned promise settle for cross-document navigations?


domenic:
Consider a case where you do

```js
appHistory.navigate("other-url").then(onFulfilled, onRejected);
```

and you have no `navigate` event handler to convert this into a same-document navigation. Here `onFulfilled` and `onRejected` will never be called, because by the time the navigate goes through, your document has been unloaded, so there's no way for `onFulfilled` and `onRejected` to run. Makes sense.

Of course, if you do that but then add a `navigate` handler which converts the navigation into a same-document navigation using `event.respondWith()`, then one of your handlers will get called, after the promise passed to `respondWith()` settles. That's the main use case for the returned promise.

Now consider the following case, again with no `navigate` event handlers:

```js
frames[0].appHistory.navigate("other-url").then(onFulfilled, onRejected);
```

Here, because we're navigating a subframe, it's very conceivable that `onFulfilled` or `onRejected` could be called, at some point after the cross-document navigation of the subframe finishes. (When, exactly? The load event? DOMContentLoaded? pageshow? readystatechange? Similar issues to #31.)

Should this work?

Our tentative answer is "no": cross-document navigations should never settle the promise. It seems like a good deal of work to implement and it's outside the usual use cases of app history which are really supposed to be about your own frame.

On the other hand, I've written a lot of test code which does the equivalent of

```js
iframe.src = "foo";
await new Promise(r => iframe.onload = r);
```

which could be nice to replace (for same-origin iframes) with

```js
await iframe.contentWindow.appHistory.navigate("foo");
```

See also https://github.com/whatwg/html/issues/5725; /cc @jakearchibald @annevk. So hmm. Maybe it would be nice.

annevk:
Unfortunately I don't have a "do this" answer, but I do have some thoughts. 😊

What happens with bfcache? Does the promise settle when you go back to the page that created it?

I'm not sure how serious @jkarlin was in https://github.com/shivanigithub/http-cache-partitioning/issues/2#issuecomment-580282052 but that was a proposal to not have `load` events for cross-origin navigations. (It's not clear if that's actually feasible though as you can still tell when `frame.contentDocument` becomes inaccessible.)

In any event, we shouldn't introduce a signal that's currently not observable, such as `DOMContentLoaded`, `pageshow`, or `readystatechange`.

jakearchibald:
[...]

Issue #94, opened 2021-04-10
Allow multiple event handlers to call `respondWith()` 


frehner:
Forked from the discussion here: https://github.com/WICG/app-history/issues/89

Some key comments from @domenic :
https://github.com/WICG/app-history/issues/89#issuecomment-814437592
https://github.com/WICG/app-history/issues/89#issuecomment-816092233

And highlights below as well

Currently only one handler can call `respondWith()`:
> To answer your specific questions:
> 
> > But what happens if an earlier event listener calls respondWith?
> 
> As currently specced, calling respondWith() or preventDefault() means that any further calls to respondWith() will fail with an "InvalidStateError" DOMException.

Some additional thoughts from @tbondwilkinson

>Of the smaller QOL improvements: I think allowing multiple calls to event.respondWith seems reasonable to me, but it begs the question whether you'd want to add the ability to get the current promise, so that you could .then() it, rather than just add to the Promise.all
>
>>Change the updating of location.href and appHistory.current to synchronously happen after all handlers have finished running (if respondWith() was called by any of them), instead of happening synchronously as part of respondWith(). This helps avoid the problem where future handlers see appHistory.current === appHistory.destination.
>
>Oh, this seems like something we'd want to do, regardless of the outcome here.

frehner:
>If we did consider the asynchronous case to be highly important, then I'd like to reconsider the decision in #46 / #19 to make all same-document navigations synchronous. In other words, if delaying action until an async determination is made is a key goal of app history, then we should just delay the whole navigation to be asynchronous, instead of doing something kludgy like telling people to synchronously intercept or cancel the navigation but later reissue it (like discussed in single-spa/single-spa#762 (comment)).

Regarding this comment, one interesting piece of feedback from #66 is that both Vue and Angular navigations are async (all Vue navigations, and Angular navigations by default). Does that feedback affect this decision?

domenic:
Not really, since they are frameworks that handle all navigations, and so they can synchronously decide to call `respondWith()`.

domenic:
Quoting https://github.com/WICG/app-history/issues/89#issuecomment-816092233:

> There are several smaller surface improvements we could make to help multiple `navigate` handlers work better together. The current design, of `respondWith()` preventing further `respondWith()` calls with no signal, is probably not optimal. Some ideas:
> 
[...]

PR #96, opened 2021-04-15
Initial spec for appHistory.navigate() and event.destination

domenic:
Regarding #95, for now the promise never settles for cross-document navigations.

Regarding #97, for now I specified `url`, `sameDocument` (not changing), and `getState()`.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/96.html" title="Last updated on Apr 23, 2021, 10:37 PM UTC (2775e31)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/96/2e154b9...2775e31.html" title="Last updated on Apr 23, 2021, 10:37 PM UTC (2775e31)">Diff</a>


spec.bs
```diff
@@ -135,12 +136,32 @@ interface AppHistory : EventTarget {
   readonly attribute boolean canGoBack;
   readonly attribute boolean canGoForward;
 
+  Promise<undefined> navigate(USVString url, optional AppHistoryNavigateOptions options = {});
+  Promise<undefined> navigate(optional AppHistoryNavigateOptions options = {});
```
domenic, line 140: Replace is not actually consulted for this overload. Perhaps we should use a fresh dictionary? This is actually observable in that the getter for `replace` will be called, whereas if we used a fresh dictionary it would not be called.

spec.bs
```diff
@@ -309,6 +308,101 @@ The following are the [=event handlers=] (and their corresponding [=event handle
...
+
+  1. <a spec="HTML" lt="parse a URL">Parse</a> |url| relative to [=this=]'s [=relevant settings object=]. If that returns failure, then return [=a promise rejected with=] a "{{SyntaxError}}" {{DOMException}}. Otherwise, let |urlRecord| be the <a spec="HTML">resulting URL record</a>.
+
+  1. Return the result of [=performing an app history navigation=] given [=this=], |urlRecord|, and |options|.
+</div>
+
+<div algorithm>
+  The <dfn method for="AppHistory">navigate(|options|)</dfn> method steps are:
+
+  1. If neither |options|["{{AppHistoryNavigationOptions/navigateInfo}}"] nor |options|["{{AppHistoryNavigateOptions/state}}"] [=map/exists=], then return [=a promise rejected with=] a {{TypeError}}.
```
domenic, line 353: This prevents `appHistory.navigate()` as some previous discussion on the issue tracker indicated that it probably indicates programmer error.

spec.bs
```diff
@@ -309,6 +308,101 @@ The following are the [=event handlers=] (and their corresponding [=event handle
...
+    <p>The default behavior of performing a full navigation to the current page can be overriden by using the {{AppHistory/navigate}} event's {{AppHistoryNavigateEvent/respondWith()}} method. Doing so will mean this call only updates state or passes along the appropriate {{AppHistoryNavigationOptions/navigateInfo}}.
+
+    <p>The returned promise behaves as described above.
+  </dd>
+</dl>
+
+<div algorithm>
+  The <dfn method for="AppHistory">navigate(|url|, |options|)</dfn> method steps are:
+
+  1. <a spec="HTML" lt="parse a URL">Parse</a> |url| relative to [=this=]'s [=relevant settings object=]. If that returns failure, then return [=a promise rejected with=] a "{{SyntaxError}}" {{DOMException}}. Otherwise, let |urlRecord| be the <a spec="HTML">resulting URL record</a>.
```
domenic, line 345: Note that the base URL chosen here is different than location.assign().

spec.bs
```diff
@@ -309,6 +308,101 @@ The following are the [=event handlers=] (and their corresponding [=event handle
...
+  1. Set |appHistory|'s [=AppHistory/navigate method call promise=] to |promise|.
+
+  1. <a spec="HTML">Navigate</a> |browsingContext| to |url| with <i>[=navigate/historyHandling=]</i> set to |historyHandling|, <i>[=navigate/appHistoryInfo=]</i> set to |navigateInfo|, and the <a spec="HTML">source browsing context</a> set to |browsingContext|.
+
+  1. Set |appHistory|'s [=AppHistory/navigate method call promise=] to |previousPromise|.
+
+  1. Return |promise|.
+</div>
+
+<p class="note">Unlike {{Location/assign()|location.assign()}} and friends, which are exposed across [=same origin-domain|origin-domain=] boundaries, {{AppHistory/navigate()|appHistory.navigate()}} can only be accessed by code with direct synchronous access to the {{Window/appHistory}} property. Thus, we avoid the complications around tracking <a spec="HTML">source browsing contexts</a>, and we don't need to deal with the <a spec="HTML">allowed to navigate</a> check and its accompanying <i>[=navigate/exceptionsEnabled=]</i> flag. We just treat all navigations as being initiated by the {{AppHistory}} object itself.
```
domenic, line 386: Note this intentional choice of source browsing context

spec.bs
```diff
@@ -466,12 +560,18 @@ An {{AppHistoryNavigateEvent}} also has an associated {{Promise}}-or-null <dfn f
      <p class="note">This can occurr if an event listener disconnected the <{iframe}> corresponding to [=this=]'s [=relevant global object=].</p>
   1. If |event|'s [=AppHistoryNavigateEvent/navigation action promise=] is non-null, or both |result| and |isSameDocument| are true, then:
     1. If |event|'s [=AppHistoryNavigateEvent/navigation action promise=] is null, then set it to [=a promise resolved with=] undefined, created in |realm|.
+    1. Let |navigateMethodCallPromise| be [=this=]'s [=AppHistory/navigate method call promise=].
```
domenic, line 565: Putting all these steps into "fire a navigate event" (instead of in the navigate algorithm itself) is starting to feel a bit strange. But it is all very app history related. We'll see how things develop as we do state and event.destination.

README.md
```diff
@@ -309,7 +309,7 @@ The event object has several useful properties:
 
 - `userInitiated`: a boolean indicating whether the navigation is user-initiated (i.e., a click on an `<a>`, or a form submission) or application-initiated (e.g. `location.href = ...`, `appHistory.navigate(...)`, etc.). Note that this will _not_ be `true` when you use mechanisms such as `button.onclick = () => appHistory.navigate(...)`; the user interaction needs to be with a real link or form. See the table in the [appendix](#appendix-types-of-navigations) for more details.
 
-- `destination`: an `AppHistoryEntry` containing the information about the destination of the navigation. Note that this entry might or might not yet be in `window.appHistory.entries()`.
+- `destination`: an object containing the information about the destination of the navigation. It has many of the same properties as an `AppHistoryEntry`: namely `url`, `sameDocument`, and `getState()` so far (but see [#97](https://github.com/WICG/app-history/issues/97) for adding more).
```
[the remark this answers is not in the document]
domenic: `sameDocument` is about whether this was *originally* a same-document navigation. But, using `respondWith()`, you can convert navigations with `sameDocument === false` into same-document navigations. In that case, the state will be applied to the (now same-document) navigation. So `getState()` should return something useful regardless.

spec.bs
```diff
@@ -309,18 +308,121 @@ The following are the [=event handlers=] (and their corresponding [=event handle
...
+
+  1. Let |promise| be [=a new promise=] created in |appHistory|'s [=relevant Realm=].
+
+  1. Let |previousPromise| be |appHistory|'s [=AppHistory/navigate method call promise=].
+
+  1. Set |appHistory|'s [=AppHistory/navigate method call promise=] to |promise|.
+
+  1. <a spec="HTML">Navigate</a> |browsingContext| to |url| with <i>[=navigate/historyHandling=]</i> set to |historyHandling|, <i>[=navigate/appHistoryInfo=]</i> set to |navigateInfo|, <i>[=navigate/appHistoryState=]</i> set to |serializedState|, and the <a spec="HTML">source browsing context</a> set to |browsingContext|.
+
+  1. Set |appHistory|'s [=AppHistory/navigate method call promise=] to |previousPromise|.
```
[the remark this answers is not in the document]
domenic: Hmm, I'll need to think more about this. It just seemed like good practice to reset it back. Let me try working through some examples.

(In simple cases this'll just reset it back to null, but in nested cases it might matter.)

spec.bs
domenic: Web IDL requires that trailing dictionaries be optional, even if logically they are not optional. See https://github.com/heycam/webidl/issues/903.

Probably we should include a comment since this is confusing to people.

spec.bs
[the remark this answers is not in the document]
domenic: I worked through the following example:

```js
  if (e.destination === url1) {
    const p2 = appHistory.navigate(url2);
  }
  e.respondWith(Promise.resolve());
});
const p1 = appHistory.navigate(url1);
```

With this spec draft `p2` and `p1` both fulfill. The URL and history update steps run first for url2 and then for url1 (because `respondWith()` happens for url2 first). This kind of makes sense. But it's not great.

I think we should tackle this at a higher level. How about the following: any call to the navigate algorithm (not just `appHistory.navigate()`, but any navigation) while a `navigate` event is ongoing forcibly cancels the ongoing navigate event, so that the new one can proceed unmolested. Probably we should also do the URL and history update steps algorithm so that `history.pushState()` during `navigate` has the same impact.

Then for the above example `p1` would reject with an "`AbortError`" `DOMException` and `p2` would fulfill. And the URL would transition to `url2` and never hit `url1`.

I'm still unsure where this leaves the original question as to whether we should restore previousPromise or not, but if this direction sounds good I can work it out.

---

A related question: consider a tweak of the above

```js
  e.respondWith(Promise.resolve());
  if (e.destination === url1) {
    const p2 = appHistory.navigate(url2);
  }
});
const p1 = appHistory.navigate(url1);
```

Here the developer has already called respondWith(). At least with the current semantics this means the url will update to url1. But probably we should fix that per https://github.com/WICG/app-history/issues/94 's "Change the updating of location.href and appHistory.current to synchronously happen after all handlers have finished running".

domenic:
Note to self: if the `navigate` event detaches the window we need to figure out what the returned promise should do. Probably forever-pending makes the most sense because any task queued to resolve the promise would be associated to a dead window/document and so would never actually happen.

domenic:
The "AbortError" cases (including ones triggered by #104) need more thinking here. Should they fire a corresponding navigateerror event? I think so; it seems simplest to keep them always in sync. I'll try to spec that out, although until #104 lands it'll be a bit messy.

natechapin:
> Note to self: if the `navigate` event detaches the window we need to figure out what the returned promise should do. Probably forever-pending makes the most sense because any task queued to resolve the promise would be associated to a dead window/document and so would never actually happen.

The behavior in https://chromium-review.googlesource.com/c/chromium/src/+/2822057 is to reject the promise, on the logic that detaching the window effectively aborts all navigations.

domenic:
Hmm, it seems we've run into https://github.com/whatwg/html/issues/2621 ...

domenic:
Ah OK, it still works. Microtasks, unlike tasks, are not document-bound.

domenic:
Great, this is rebased on top of all the somewhat-related changes and is ready for re-review.

natechapin (approved):

PR #103, opened 2021-04-23
Move URL/history updating out of respondWith()

domenic:
This ensures that if multiple event handlers are in place, appHistory.current and location.href only update after they all have finished.

Part of #94.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/103.html" title="Last updated on Apr 23, 2021, 5:55 PM UTC (942f3f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/103/e2d9f6b...942f3f9.html" title="Last updated on Apr 23, 2021, 5:55 PM UTC (942f3f9)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -423,11 +423,6 @@ An {{AppHistoryNavigateEvent}} also has an associated {{Promise}}-or-null <dfn f
   1. If [=this=]'s [=Event/canceled flag=] is set, then throw an "{{InvalidStateError}}" {{DOMException}}.
   1. Set [=this=]'s [=Event/canceled flag=].
   1. Set [=this=]'s [=AppHistoryNavigateEvent/navigation action promise=] to |newNavigationAction|.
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -464,6 +459,12 @@ An {{AppHistoryNavigateEvent}} also has an associated {{Promise}}-or-null <dfn f
   1. Let |result| be the result of [=dispatching=] |event| at |appHistory|.
   1. If [=this=]'s [=relevant global object=]'s [=Window/browsing context=] is null, then return false.
      <p class="note">This can occurr if an event listener disconnected the <{iframe}> corresponding to [=this=]'s [=relevant global object=].</p>
+  1. If |event|'s [=AppHistoryNavigateEvent/navigation action promise=] is non-null, then:
```
natechapin, line 462: `navigation action promise` will only be non-null here if `respondeWith()` was called, right?

natechapin (approved):

spec.bs
domenic: Exactly, it's basically synonymous with that.

PR #109, opened 2021-05-07
Specify traversal and overhaul ongoing navigation tracking

domenic:
This commit introduces a much more robust framework for tracking ongoing navigations. This fixes #130, for example, by appropriately keeping the signal around so that we can signal abort when necessary. It also ensures every same-document navigation ends in either navigationsuccess or navigationerror, with a few unified spec paths.

With this framework in hand, we can specify goTo()/back()/forward().

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/109.html" title="Last updated on Jul 27, 2021, 9:30 PM UTC (c194cdd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/109/6277f4b...c194cdd.html" title="Last updated on Jul 27, 2021, 9:30 PM UTC (c194cdd)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -58,10 +58,7 @@ spec: html; urlPrefix: https://whatpr.org/html/6315/
     text: get all history steps; for: traversable navigable; url: history.html#getting-all-history-steps
     text: step; for: session history entry; url: history.html#she-step
     text: apply the history step; url: history.html#apply-the-history-step
-    text: checkForUserCancelation; for: apply the history step; url: history.html#apply-the-history-step
-    text: initiatorToCheck; for: apply the history step; url: history.html#apply-the-history-step
     text: containing navigable; for: browsing context; url: browsers.html#bc-navigable
-    text: session; for: browsing context group; url: browsers.html#bcg-session
     text: active document; for: navigable; url: history.html#nav-document
     text: navigable; url: history.html#navigable
 spec: uuid; type: dfn; urlPrefix: https://wicg.github.io/uuid/
@@ -157,6 +154,10 @@ interface AppHistory : EventTarget {
   Promise<undefined> navigate(USVString url, optional AppHistoryNavigateOptions options = {});
   Promise<undefined> reload(optional AppHistoryReloadOptions options = {});
 
+  Promise<undefined> goTo(DOMString key, optional AppHistoryNavigationOptions options = {});
+  Promise<undefined> back(optional AppHistoryNavigationOptions options = {});
+  Promise<undefined> forward(optional AppHistoryNavigationOptions options = {});
+
   attribute EventHandler onnavigate;
   attribute EventHandler onnavigatesuccess;
   attribute EventHandler onnavigateerror;
@@ -175,16 +176,6 @@ dictionary AppHistoryReloadOptions : AppHistoryNavigationOptions {
   any state;
 };
 </xmp>
-
-Each {{AppHistory}} object has an associated <dfn for="AppHistory">entry list</dfn>, a [=list=] of {{AppHistoryEntry}} objects, initially empty.
-
-Each {{AppHistory}} object has an associated <dfn for="AppHistory">current index</dfn>, an integer, initially &minus;1.
-
-Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing navigate event</dfn>, an {{AppHistoryNavigateEvent}} or null, initially null.
-
-Each {{AppHistory}} object has an associated <dfn for="AppHistory">navigate method call promise</dfn>, which is either a {{Promise}} or null, initially null.
-
-Each {{AppHistory}} object has an associated <dfn for="AppHistory">navigate method call serialized state</dfn>, which is either a [=serialized state=] or null, initially null.
 
 <h3 id="entries-api">Introspecting the app history entry list</h3>
 
@@ -209,6 +200,10 @@ Each {{AppHistory}} object has an associated <dfn for="AppHistory">navigate meth
     <p>Returns true if the current {{AppHistoryEntry}} is not the last one in the app history entries list.
   </dd>
 </dl>
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">entry list</dfn>, a [=list=] of {{AppHistoryEntry}} objects, initially empty.
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">current index</dfn>, an integer, initially &minus;1.
 
 <div algorithm>
   The <dfn attribute for="AppHistory">current</dfn> getter steps are:
@@ -346,6 +341,143 @@ Each {{AppHistory}} object has an associated <dfn for="AppHistory">navigate meth
   1. Assert: this step is never reached.
 </div>
 
+<h3 id="ongoing-state">Ongoing navigation tracking</h3>
+
+During any given navigation, the {{AppHistory}} object needs to keep track of the following:
+
+<table class="data">
+  <caption>For non-"{{AppHistoryNavigationType/traverse}}" navigations
+  <thead>
+    <tr>
+      <th>State
+      <th>Duration
+      <th>Explanation
+  <tbody>
+    <tr>
+      <td>The {{AppHistoryNavigateEvent}}
+      <td>For the duration of event firing
+      <td>So that if the navigation is canceled while the event is firing, we can [=Event/canceled flag|cancel=] the event.
+    <tr>
+      <td>Any {{AppHistoryNavigateOptions/state}}
+      <td>For the duration of event firing
+      <td>So that we can update the current entry's state after the event successfully finishes firing without being canceled.
+    <tr>
+      <td>The event's {{AppHistoryNavigateEvent/signal}}
+      <td>Until all promises passed to {{AppHistoryNavigateEvent/respondWith()}} have settled
+      <td>So that if the navigation is canceled, we can [=AbortSignal/signal abort=].
+    <tr>
+      <td>Any {{Promise}} that was returned
+      <td>Until all promises passed to {{AppHistoryNavigateEvent/respondWith()}} have settled
+      <td>So that we can [=resolve=] or [=reject=] it appropriately.
+</table>
+
+<table class="data">
+  <caption>For "{{AppHistoryNavigationType/traverse}}" navigations
+  <thead>
+    <tr>
+      <th>State
+      <th>Duration
+      <th>Explanation
+  <tbody>
+    <tr>
+      <td>The {{AppHistoryNavigateEvent}}
+      <td>For the duration of event firing
+      <td>So that if the navigation is canceled while the event is firing, we can [=Event/canceled flag|cancel=] the event.
+    <tr>
+      <td>Any {{AppHistoryNavigationOptions/navigateInfo}}
+      <td>Until the task is queued to fire the {{AppHistory/navigate}} event
+      <td>So that we can use it to fire the {{AppHistory/navigate}} event after the the trip through the [=traversable navigable/session history traversal queue=]
+    <tr>
+      <td>The event's {{AppHistoryNavigateEvent/signal}}
+      <td>Until all promises passed to {{AppHistoryNavigateEvent/respondWith()}} have settled
+      <td>So that if the navigation is canceled, we can [=AbortSignal/signal abort=].
+    <tr>
+      <td>Any {{Promise}} that was returned
+      <td>Until all promises passed to {{AppHistoryNavigateEvent/respondWith()}} have settled
+      <td>So that we can [=resolve=] or [=reject=] it appropriately.
+</table>
+
+Furthermore, we need to account for the fact that there might be multiple traversals queued up, e.g. via
+
+<xmp highlight="js">
+const key1 = appHistory.entries()[appHistory.current.index - 1].key;
+const key2 = appHistory.entries()[appHistory.current.index + 1].key;
+
+appHistory.goTo(key1); // intentionally no await
+appHistory.goTo(key2);
+</xmp>
+
+And, while non-traversal navigations cannot be queued in the same way since a new non-traversal navigation cancels an old one, we need to keep some state around so that we can properly cancel the old one. That is, given
+
+<xmp highlight="js">
+const p1 = appHistory.navigate(url1);
+const p2 = appHistory.navigate(url2);
+</xmp>
+
+we need to ensure that when navigating to `url2`, we still have the {{Promise}} `p1` around so that we can reject it. We can't just get rid of any ongoing navigation promises the moment the second call to {{AppHistory/navigate()}} happens.
+
+We end up accomplishing all this using the following setup:
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing navigate event</dfn>, an {{AppHistoryNavigateEvent}} or null, initially null.
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">to-be-set serialized state</dfn>, a [=serialized state=] or null, initially null.
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">upcoming non-traverse navigation</dfn>, which is an [=app history API navigation=] or null, initially null.
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing non-traverse navigation</dfn>, which is an [=app history API navigation=] or null, initially null.
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing traverse navigations</dfn>, which is a [=map=] from strings to [=app history API navigations=], initially empty.
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">post-navigate event ongoing navigation signal</dfn>, which is an {{AbortSignal}} or null, initially null.
+
+An <dfn>app history API navigation</dfn> is a [=struct=] with the following [=struct/items=]:
+
+* An <dfn for="app history API navigation">info</dfn>, a JavaScript value
+* A <dfn for="app history API navigation">returned promise</dfn>, a {{Promise}} or null
+* A <dfn for="app history API navigation">cleanup step</dfn>, an algorithm step
+
+<p class="note">We need to store the {{AbortSignal}} separately from the [=app history API navigation=] struct, since it needs to be tracked even for navigations that are not via the app history APIs. So, we store it some of the time in the [=AppHistory/ongoing navigate event=]'s {{AppHistoryNavigateEvent/signal}} property, and the rest of the time in the [=AppHistory/post-navigate event ongoing navigation signal=].
+
+<div algorithm>
+  To <dfn for="AppHistory">set the upcoming non-traverse navigation</dfn> given an {{AppHistory}} |appHistory| and a JavaScript value |info|:
+
+  1. Let |promise| be [=a new promise=] created in |appHistory|'s [=relevant Realm=].
+
+  1. Let |cleanupStep| be an algorithm step which sets |appHistory|'s [=AppHistory/ongoing non-traverse navigation=] to null.
+
+  1. Let |ongoingNavigation| be an [=app history API navigation=] whose [=app history API navigation/info=] is |info|, [=app history API navigation/returned promise=] is |promise|, and [=app history API navigation/cleanup step=] is |cleanupStep|.
+
+  1. Assert: |appHistory|'s [=AppHistory/upcoming non-traverse navigation=] is null.
+
+  1. Set |appHistory|'s [=AppHistory/upcoming non-traverse navigation=] to |ongoingNavigation|.
+
+  1. Return |ongoingNavigation|.
+</div>
+
+<div algorithm>
+  To <dfn for="AppHistory">promote the upcoming non-traverse navigation to ongoing</dfn> given an {{AppHistory}} |appHistory|:
+
+  1. Assert: |appHistory|'s [=AppHistory/ongoing non-traverse navigation=] is null.
+
+  1. Set |appHistory|'s [=AppHistory/ongoing non-traverse navigation=] to |appHistory|'s [=AppHistory/upcoming non-traverse navigation=].
+
+  1. Set |appHistory|'s [=AppHistory/upcoming non-traverse navigation=] to null.
+</div>
+
+<div algorithm>
+  To <dfn for="AppHistory">set an ongoing traverse navigation</dfn> given an {{AppHistory}} |appHistory|, a string |key|, and a JavaScript value |info|:
+
+  1. Let |promise| be [=a new promise=] created in |appHistory|'s [=relevant Realm=].
+
+  1. Let |cleanupStep| be an algorithm step which [=map/removes=] |appHistory|'s [=AppHistory/ongoing traverse navigations=][|key|].
+
+  1. Let |ongoingNavigation| be an [=app history API navigation=] whose [=app history API navigation/info=] is |info|, [=app history API navigation/returned promise=] is |promise|, and [=app history API navigation/cleanup step=] is |cleanupStep|.
+
+  1. Set |appHistory|'s [=AppHistory/ongoing traverse navigations=][|key|]  to |ongoingNavigation|.
+
+  1. Return |ongoingNavigation|.
+</div>
+
 <h3 id="global-navigate">Navigating</h3>
 
 <dl class="domintro non-normative">
@@ -399,7 +531,7 @@ Each {{AppHistory}} object has an associated <dfn for="AppHistory">navigate meth
 
   1. Let |historyHandling| be "<a for="history handling behavior">`replace`</a>" if |options|["{{AppHistoryNavigateOptions/replace}}"] is true; otherwise, "<a for="history handling behavior">`default`</a>".
 
-  1. Return the result of [=performing an app history navigation=] given [=this=], |urlRecord|, |serializedState|, |navigateInfo|, and |historyHandling|.
+  1. Return the result of [=performing a non-traverse app history navigation=] given [=this=], |urlRecord|, |serializedState|, |navigateInfo|, and |historyHandling|.
 </div>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -403,6 +407,101 @@ Each {{AppHistory}} object has an associated <dfn for="AppHistory">navigate meth
...
+
+  1. If |appHistory|'s [=relevant global object=]'s [=associated Document=] is not [=Document/fully active=], then return [=a promise rejected with=] an "{{InvalidStateError}}" {{DOMException}}.
+
+  1. Let |jsh| be |appHistory|'s [=relevant global object=]'s [=Window/browsing context=]'s [=top-level browsing context=]'s <a spec="HTML">joint session history</a>.
+
+  1. Let |currentEntryIndex| be the index of the <a spec="HTML" lt="current entry of the joint session history">current entry</a> within |jsh|.
+
+  1. Let |destinationIndex| be the index of the closest [=session history entry=] to the <a spec="HTML" lt="current entry of the joint session history">current entry</a>, such that either that entry's [=session history entry/app history key=] equals |key|, or that entry is "standing in place of" another entry whose [=session history entry/app history key=] equals |key|. Here, |entryA| "standing in place of" <var ignore>entryB</var> means that they occupy the same "slot" within the joint session history, but only |entryA| is included because it most recently became the [=session history/current entry=] of its [=session history=].
```
domenic, line 494: Actually, maybe the "standing in place of" thing is unnecessary, because you can only perform a navigation which navigates at least your frame (and potentially child frames)?

So maybe the algorithm could just look through your browsing context's session history for the one with the right key, and then assert that that SHE exists in the joint session history?

I'm not sure how worthwhile it is to polish this step given that the infrastructure is going to change a lot, but I couldn't stop thinking about it...

domenic:
Per #110 I don't think `appHistory.goTo()`/`back()`/`forward()` will ever trigger the "allowed to navigate" check. So instead the potential error cases are unload/beforeunload/user refuses to allow the document to be unloaded.

domenic:
> Probably navigate() should have the same "bails out during unload" behavior.

I'll block this on https://github.com/whatwg/html/pull/5666

spec.bs
domenic: This is wrong as shown by the example in https://github.com/WICG/app-history/issues/110#issuecomment-834831307

jakearchibald:
What would you expect `appHistory.forward(); appHistory.back()` to do? Right now those methods look up state synchronously (current index), but ultimately navigations queue in some way.

`appHistory.forward()` also looks up the entry list size, which may be incorrect (in the same way `history.length` can be incorrect). We can try to keep it up to date like `history.length`, but maybe an async approach is better?

How I see traversal working with https://github.com/whatwg/html/pull/6315:

1. Let navigable be the current document's navigable.
1. Let traversable be navigable's [traversable](https://whatpr.org/html/6315/history.html#nav-traversable).
1. Queue the rest of the steps on that traversable navigable's [session history traversal queue](https://whatpr.org/html/6315/history.html#tn-session-history-traversal-queue):
1. Let `targetStep` be the result of getting the previous/next/goTo step given the current navigable (see below).
1. If `targetStep` is a failure, reject promise.
1. [Apply the history step](https://whatpr.org/html/6315/history.html#apply-the-history-step) `targetStep` with checkForUserCancelation set to true and initiatorToCheck set to document's browsing context.

There are a few ways this can 'fail', and I'm not sure what you'd like to do for each of these cases:

- The current browsing context is not allowed to navigate one of the navigables involved in this traversal. This causes the whole traversal to be abandoned.
- One of the documents-to-unload asks for a beforeunload prompt to be shown, and the user presses cancel. This causes the whole traversal to be abandoned.
- The traversal is cancelled in favour of another traversal, or the user presses 'stop' (I haven't factored this into the spec yet, but the current spec has it in a hand-wavey and incorrect way)
- The network returns a `Content-Disposition`, 204, or 205 when fetching the document for the target entry for this navigable. This will result in the current navigable sticking with the same active session history item, but other navigables may successfully navigate as part of the traversal.
- Other navigables involved in the traversal stick with their current active entry due to a `Content-Disposition`, 204, or 205 response.
- Other navigables involved in the traversal display an error document due to some other fetch failure.

Getting the previous step given the current navigable:

1. Let currentEntry be navigable's [active session history entry](https://whatpr.org/html/6315/history.html#nav-active-history-entry).
1. Let navigableEntries be navigable's [session history entries](https://whatpr.org/html/6315/history.html#getting-session-history-entries).
1. Filter navigableEntries so it contains contiguous entries that are part of the same origin and [session](https://whatpr.org/html/6315/browsers.html#bcg-session). (I don't have an algorithm for this right now)
1. If navigableEntries[0] is currentEntry, return failure.
1. Let allSteps be [all history steps](https://whatpr.org/html/6315/history.html#getting-all-history-steps) in navigable's [traversable navigable](https://whatpr.org/html/6315/history.html#nav-traversable) within navigable's BCG's [session](https://whatpr.org/html/6315/browsers.html#bcg-session).
1. Return the greatest number in allSteps that is less than currentEntry's [step](https://whatpr.org/html/6315/history.html#she-step). This ensures we're going the minimum number of steps back to get to the previous entry for this navigable.

I'm not 100% sure if we should be using the [active session history entry](https://whatpr.org/html/6315/history.html#nav-active-history-entry) or the [current session history entry](https://whatpr.org/html/6315/history.html#nav-current-history-entry). It kinda depends on what you expect to happen if `back()` is called, but the result is a 204, then `back()` is called again. If a 204 is returned, the 'current' history entry changes, but the 'active' entry stays the same.

Getting the next step given the current navigable:

1. Let currentEntry be navigable's [active session history entry](https://whatpr.org/html/6315/history.html#nav-active-history-entry).
1. Let navigableEntries be navigable's [session history entries](https://whatpr.org/html/6315/history.html#getting-session-history-entries).
1. Filter navigableEntries so it contains contiguous entries that are part of the same origin and [session](https://whatpr.org/html/6315/browsers.html#bcg-session). (I don't have an algorithm for this right now)
1. If navigableEntries[navigableEntries.length - 1] is currentEntry, return failure.
1. Return the [step](https://whatpr.org/html/6315/history.html#she-step) of the item in navigableEntries that appears after currentEntry.

Getting the `goTo` step given the current navigable and a key:

1. Let currentEntry be navigable's [active session history entry](https://whatpr.org/html/6315/history.html#nav-active-history-entry).
1. Let navigableEntries be navigable's [session history entries](https://whatpr.org/html/6315/history.html#getting-session-history-entries).
1. Filter navigableEntries so it contains contiguous entries that are part of the same origin and [session](https://whatpr.org/html/6315/browsers.html#bcg-session). (I don't have an algorithm for this right now)
1. Let targetEntry be the entry in navigableEntries that has key == key (I guess this is where we'll put it)
1. If targetEntry is currentEntry ???
1. Return targetEntry's [step](https://whatpr.org/html/6315/history.html#she-step).

domenic:
> What would you expect appHistory.forward(); appHistory.back() to do? Right now those methods look up state synchronously (current index), but ultimately navigations queue in some way.

Good question. The second call should cancel the first (probably using something like the current spec's "Remove any tasks queued by the history traversal task source that are associated...") and thus win. And then we should propagate the failure back to the return value of the first call.

> appHistory.forward() also looks up the entry list size, which may be incorrect (in the same way history.length can be incorrect). We can try to keep it up to date like history.length, but maybe an async approach is better?

I think we should do both. If the synchronously-accessible `appHistory.canGoForward`/`appHistory.canGoBack` are false, then `appHistory.forward()`/`appHistory.back()` should fail immediately. But yeah, they might also fail asynchronously.

> How I see traversal working with whatwg/html#6315:

Thank you so much for outlining this and the accompanying algorithms!! I think I will try to just copy them in, linking to your draft PR. Trying to do anything rigorous given the current spec is an exercise in futility IMO.

> The current browsing context is not allowed to navigate one of the navigables involved in this traversal. This causes the whole traversal to be abandoned.

Reject promise with "SecurityError" DOMException.

> One of the documents-to-unload asks for a beforeunload prompt to be shown, and the user presses cancel. This causes the whole traversal to be abandoned.

Reject promise with an "AbortError" DOMException.

> The traversal is cancelled in favour of another traversal, or the user presses 'stop' (I haven't factored this into the spec yet, but the current spec has it in a hand-wavey and incorrect way)

Reject promise with an "AbortError" DOMException.

> The network returns a Content-Disposition, 204, or 205 when fetching the document for the target entry for this navigable. This will result in the current navigable sticking with the same active session history item, but other navigables may successfully navigate as part of the traversal.

Probably treat this as a success (see #95 for what that means). Especially since, IIUC, the plan is for this to update the "current" entry, even if it doesn't update the "active" one?

> Other navigables involved in the traversal stick with their current active entry due to a Content-Disposition, 204, or 205 response.
>
> Other navigables involved in the traversal display an error document due to some other fetch failure.

Definitely a success.

jakearchibald:
> > What would you expect appHistory.forward(); appHistory.back() to do? Right now those methods look up state synchronously (current index), but ultimately navigations queue in some way.
> 
> Good question. The second call should cancel the first (probably using something like the current spec's "Remove any tasks queued by the history traversal task source that are associated...") and thus win. And then we should propagate the failure back to the return value of the first call.

That makes as much sense as anything. Ugh, I just gave some of this a quick test:

- `history.back(); history.back()` - Seems to always go(-2)
- `history.forward(); history.forward()` - Seems to always go(2)
- `history.back(); history.forward()` - Seems to always go(-1)
- `history.forward(); history.back()` - no-op if forward navigation is another page. go(1) if forward navigation is same document.

I'm sure there are more subtleties, and I only tested Chrome. Pretty sure this won't be interoperable.

> > appHistory.forward() also looks up the entry list size, which may be incorrect (in the same way history.length can be incorrect). We can try to keep it up to date like history.length, but maybe an async approach is better?
> 
> I think we should do both. If the synchronously-accessible `appHistory.canGoForward`/`appHistory.canGoBack` are false, then `appHistory.forward()`/`appHistory.back()` should fail immediately. But yeah, they might also fail asynchronously.

I can't tempt you to make `canGoForward`/`canGoBack` async? 😄 

But yeah, this won't be hard to spec, as the hooks are already there for history length and index.

> Thank you so much for outlining this and the accompanying algorithms!!

It was a good test of the central "go here" algorithm!

> > The network returns a Content-Disposition, 204, or 205 when fetching the document for the target entry for this navigable. This will result in the current navigable sticking with the same active session history item, but other navigables may successfully navigate as part of the traversal.
> 
> Probably treat this as a success (see #95 for what that means). Especially since, IIUC, the plan is for this to update the "current" entry, even if it doesn't update the "active" one?

Yeah, that's correct. I think "success" here is reasonable, even though it might be weird that a cross-document navigation is a success without changing page.

If we want to avoid this edge case, we can introduce a new traversal mode that creates an error document in these cases, so the page will change.

This isn't a problem for navigations that create a new history entry, or reloads, as the network stuff happens before trying to modify the history tree, and if the result is a 204/5/`content-disposition` then the history tree isn't modified at all.

spec.bs
```diff
@@ -403,6 +409,107 @@ Each {{AppHistory}} object has an associated <dfn for="AppHistory">navigate meth
...
+    * For cross-document traversals, it will never settle.
+  </dd>
+</dl>
+
+<div algorithm>
+  The <dfn method for="AppHistory">goTo(|key|, |options|)</dfn> method steps are:
+
+  1. If [=this=]'s [=AppHistory/current index=] is &minus;1, then return [=a promise rejected with=] an "{{InvalidStateError}}" {{DOMException}}.
+
+  1. If [=this=]'s [=AppHistory/entry list=][[=this=]'s [=AppHistory/current index=]]'s [=AppHistoryEntry/session history entry=]'s [=session history entry/app history key=] equals |key|, then return [=a promise resolved with=] undefined.
```
domenic, line 460: This is not correct; it needs to check all the error conditions first. So this should move to "performing an app history traversal".

domenic:
Wow, the current spec draft doesn't actually return a promise for the success case. No wonder it's relatively simple. It needs to do similar things to how navigate() works, with "fire a navigate event" dealing with the promise.

domenic:
Some easy questions:

> Let navigable be the current document's navigable.

Is this pointer defined anywhere in the draft?

> Let targetStep be the result of getting the previous/next/goTo step given the current navigable (see below).

By this you mean _navigable_, not _traversable_?

Trickier question:

> Getting the previous step given the current navigable:
>
> Getting the next step given the current navigable:
>
> Getting the goTo step given the current navigable and a key:

I was thinking we could unify these into just the goTo algorithm, since we already know what the next/previous key are. I don't believe it's possible for the notion of next/previous key to get out of sync, because anything that affects the next/previous key would have to navigate the current frame and we'd know about it.

Similarly,

> Filter navigableEntries so it contains contiguous entries that are part of the same origin and session. (I don't have an algorithm for this right now)

I'm not sure the filtering is necessary since we already filtered the list of possible keys?

domenic:
> Let currentEntry be navigable's active session history entry.

Should this be current or active? I think current?

domenic:
I'm realizing that I'm going to need an updated version of "traverse the history by a delta" in order to specify `navigate` events properly for `history.back()`/etc., and thus make sure I also specify them in a way that works for `appHistory.back()`/etc.

Would you be able to prioritize

> JAKE-TODO: this needs to be rewritten to figure out the target step and run "apply the history step" 

? I gave it a shot myself, but I'm unsure about the deltas of translating a delta into a step.

jakearchibald:
> > Let navigable be the current document's navigable.
> 
> Is this pointer defined anywhere in the draft?

There's `<span data-x="bc-navigable">containing navigable</span>`, so you can go via the browsing context.

> > Let targetStep be the result of getting the previous/next/goTo step given the current navigable (see below).
> 
> By this you mean _navigable_, not _traversable_?

Yes. It has to be the navigable, since app history back/forward is specific to the navigable. `history.back`/`forward` could just target the traversable.

> > Getting the previous step given the current navigable:
> > Getting the next step given the current navigable:
> > Getting the goTo step given the current navigable and a key:
> 
> I was thinking we could unify these into just the goTo algorithm, since we already know what the next/previous key are.

I think it might be more complicated. For example:

1. Navigate to page with 2 iframes - All three history entries are step 1
1. Navigate iframe-1 to `/foo` - step 2
1. Navigate iframe-2 to `/bar` - step 3
1. Navigate iframe-1 to `/hello` - step 4

If iframe-1 calls back, I think we need to go to step 3, right? The idea is to traverse the minimum number of steps to activate that history entry, rather than go to the step where that history entry was created. Hmm, that means my `goTo` algorithm is wrong, although my `back` algorithm is right.

> > Filter navigableEntries so it contains contiguous entries that are part of the same origin and session. (I don't have an algorithm for this right now)
> 
> I'm not sure the filtering is necessary since we already filtered the list of possible keys?

Hmm, I'm not sure. The `navigableEntries` are filtered by origin, but the steps are only filtered by session. The steps are overall steps in the traversable, which will include other iframe and the top level, whereas the `navigableEntries` are the once the app history API has access to.

> > Let currentEntry be navigable's active session history entry.
> 
> Should this be current or active? I think current?

I'm still not 100% sure what to do here. But yeah, let's go for current for now.

> Would you be able to prioritize
> 
> > JAKE-TODO: this needs to be rewritten to figure out the target step and run "apply the history step"

Yeah, shall do

jakearchibald:
> > Let currentEntry be navigable's active session history entry.
> 
> Should this be current or active? I think current?

Ok, I've thought about this some more, and I think this should be "active". In fact, I might even remove the concept of "current", since I think it's risky and can be replaced with something else.

Take the following history timeline:

Screenshot 2021-05-17 at 15 07 40

If we're on step 4, and want to go `back()`, but `/foo` returns a `Content-Disposition`/204/205, we still move the history pointer to step 3. This is done to avoid trapping the user in one session history entry.

However, when we go `back()` again, to step 2, only the second iframe should try to navigate. My initial spec was along the lines of "if your destination history entry is different to your active history entry, then you need to navigate", but that would try to navigate the first iframe again too, so I created the split between "current" and "active" history entries. However, I think I can achieve this by replacing "current history entry" with "get the history entry associated with the current history step", along with some warning saying "you probably want the active history entry".

For example:

Screenshot 2021-05-17 at 15 20 46

If we're on step 2, and go `back()`, but `//example.com/` returns a 204, you wouldn't expect `history.state` to be `{foo: 'bar'}`, as that would be leaking data across origins. That's what would happen if `history.state` came from the "current history entry".

app-history doesn't have this risk, as it's limited to the current frame and can't cross origins, but it seems best for consistency that app-history gets its state from the active history entry. Otherwise, you can end up with "current" state that was never associated with the active document.

That raises questions over how `appHistory.back()` etc should behave:

Screenshot 2021-05-17 at 15 41 14

Let's say we're on step 3, but we're still showing `/hello` because `/bar` returned a 204.

`history.back()` operates like the back button, so it seems obvious that this should traverse to step 2 (`/foo`), even though, prior to calling `history.back()`, `history.state` would be whatever's associated with the entry in step 4. Essentially, `history.back()` decreases the step number by 1 (although in reality it'll skip steps that no longer have entries associated with them, due to removed iframes).

This isn't so clear with `appHistory.back()`, which deliberately scopes its view to a single navigable. Options:

- Apply step 2. This means going back two entries compared to what's in `appHistory.current`, which is kinda weird.
- Stay on step 3, but re-attempt to make `/bar` the active entry by requesting `/bar` again, which may result in another download/204/205.

I think the latter is more consistent with what app-history is trying to do. Maybe this also means that `appHistory.back()` should reject if the result is an unchanged `appHistory.current`, even if it successfully changes the history entry in other navigables.

domenic:
> If iframe-1 calls back, I think we need to go to step 3, right? The idea is to traverse the minimum number of steps to activate that history entry, rather than go to the step where that history entry was created. Hmm, that means my goTo algorithm is wrong, although my back algorithm is right.

Correct, the intention would be to go to step 3. So I guess the key is to realize that a key uniquely identifies a session history entry, but does not uniquely identify a joint session history step. Makes sense.

I still think we should spec goTo and then spec back/forward in terms of those, though, if possible. I guess goTo would, given _key_:

- Find all steps which contain the entry with key _key_
- Pick the step _S_ such that |_S_ - _currentStep_| is smallest.

> Hmm, I'm not sure. The navigableEntries are filtered by origin, but the steps are only filtered by session. The steps are overall steps in the traversable, which will include other iframe and the top level, whereas the navigableEntries are the once the app history API has access to.

Is the worry that, given a key mapping to multiple eligible steps, the closest eligible step might be... cross-origin? Different session? I'm not sure what the problem is... An example case would help (and we could convert it into a web platform test), like the one for the previous question which set me straight.

> Ok, I've thought about this some more, and I think this should be "active". In fact, I might even remove the concept of "current", since I think it's risky and can be replaced with something else.

Hmm. So in your proposal, what is "get the history entry associated with the current history step" used for and what is active used for? From what I can tell:

- current = joint session history backs (e.g. `history.back()` and the user pressing the reload button?). Maybe reloads?
- active = `history.state`, `appHistory.current`, `appHistory.back()`... Also, it'll stay in sync with `document.URL`, although somewhat by coincidence.

If we do that, I agree "Stay on step 3, but re-attempt to make /bar the active entry by requesting /bar again, which may result in another download/204/205" seems like the natural behavior: it'll just try to re-traverse to the `key` which is previous to `appHistory.current`, which is consistent with my general desire to spec `back()` in terms of `goTo()`.

> Maybe this also means that appHistory.back() should reject if the result is an unchanged appHistory.current, even if it successfully changes the history entry in other navigables.

This would be ideal, I agree.

jakearchibald:
> I still think we should spec goTo and then spec back/forward in terms of those, though, if possible. I guess goTo would, given _key_:
> 
> * Find all steps which contain the entry with key _key_
> * Pick the step _S_ such that |_S_ - _currentStep_| is smallest.

Taking another swing at it: Getting the `goTo` step given the current navigable and a key:

1. Let traversable be navigable's [traversable](https://whatpr.org/html/6315/history.html#nav-traversable).
1. Let activeEntry be navigable's [active session history entry](https://whatpr.org/html/6315/history.html#nav-active-history-entry).
1. Let navigableEntries be navigable's [session history entries](https://whatpr.org/html/6315/history.html#getting-session-history-entries).
1. Filter navigableEntries so it contains contiguous entries that are part of the same origin and [session](https://whatpr.org/html/6315/browsers.html#bcg-session). (I don't have an algorithm for this right now).
1. Let targetEntry be the entry in navigableEntries that has key == key.
1. If targetEntry is activeEntry ???.
1. If targetEntry's [step](https://whatpr.org/html/6315/history.html#she-step) is greater than traversable's [step](https://whatpr.org/html/6315/history.html#tn-current-session-history-step), return targetEntry's [step](https://whatpr.org/html/6315/history.html#she-step).
1. Let afterTarget be the item after targetEntry in navigableEntries.
1. Let allSteps be [all history steps](https://whatpr.org/html/6315/history.html#getting-all-history-steps) in traversable within navigable's BCG's [session](https://whatpr.org/html/6315/browsers.html#bcg-session).
1. Return the greatest number in allSteps that is less than afterTarget's [step](https://whatpr.org/html/6315/history.html#she-step).

> Is the worry that, given a key mapping to multiple eligible steps, the closest eligible step might be... cross-origin? Different session? I'm not sure what the problem is... An example case would help (and we could convert it into a web platform test), like the one for the previous question which set me straight.

I might be getting confused. Here's where my mind's at:

Screenshot 2021-05-19 at 12 47 02

Here we've got 7 history entries (two of them in an iframe), 6 steps, and 5 entries that are linked to the top level navigable.

All steps are part of the same session. The session only changes if the user performs a manual cross-origin navigation outside the page.

If we're on step 6, the top level can only go to `k3`, `k4` and `k7` (which is the current step). Because:

- `k5` and `k6` are associated with a different navigable (although `k5` is on the same step as `k4`).
- `k1` and `k2` are part of the same navigable, but they're not same-origin-contiguous with the current entry, even though `k1` is same origin.

To enforce this it seems best to filter the history entries associated with the current navigable given a current step. That returns the entries with keys `k3`, `k4` and `k7`.

When going from step 6 to key `k4` we actually want to go to step 5, since this is the minimum movement. To do this, we:

1. Let targetEntry be the entry in navigableEntries that has key == `k4`.
1. Let afterTarget be the item after targetEntry in navigableEntries. This is the item with key `k7` and step 6.
1. Let allSteps be [all history steps](https://whatpr.org/html/6315/history.html#getting-all-history-steps) in traversable within navigable's BCG's [session](https://whatpr.org/html/6315/browsers.html#bcg-session). This is 1-6 inclusive.
1. Return the greatest number in allSteps that is less than afterTarget's [step](https://whatpr.org/html/6315/history.html#she-step) (6). So we return 5.

Filtering all history steps to a particular session probably isn't required, but I'd like to make the session a required argument for security reasons. Only the browser back/forward button should be able to bypass the session check.

> Hmm. So in your proposal, what is "get the history entry associated with the current history step" used for and what is active used for? From what I can tell:
> 
> * current = joint session history backs (e.g. `history.back()` and the user pressing the reload button?). Maybe reloads?
> * active = `history.state`, `appHistory.current`, `appHistory.back()`... Also, it'll stay in sync with `document.URL`, although somewhat by coincidence.

Yeah, that looks right to me. I'd need to do some tests of reload to be sure.

domenic:
> If we're on step 6, the top level can only go to k3, k4 and k7 (which is the current step). Because:

Agreed. However, `appHistory.entries().map(e => e.key)` only contains k3, k4, and k7. I.e. the filtering has already been done. So I don't think we need to filter in the traversal step.

We do need to check that the key we're given is in (the spec equivalent of) `appHistory.entries().map(e => e.key)`, since otherwise a cooperating iframe could give you an invalid key. But that check can be done synchronously.

Now, in implementations, probably we want to do the check in the browser process for security reasons... and maybe the spec should reflect that too. But I'm not sure it's necessary.

domenic:
> navigable's BCG's session.

I don't see a definition for navigable's BCG in https://whatpr.org/html/6315/history.html#navigables ... should there be one?

domenic:
> If targetEntry's step is greater than traversable's step, return targetEntry's step.

traversable's step is currently named "current session history step". Naming-wise, should that change to "active"? Is it equivalent to/redundant with the navigable's active session history entry's step?

In particular:

> Let afterTarget be the item after targetEntry in navigableEntries.

I think you're assuming this always exists because you're handling three cases:

- targetEntry step > "current step": return targetEntry's step
- targetEntry step = "current step": ???
- targetEntry step < "current step": do more complicated dance involving afterTarget

But this only works if the same notion of "current step" is used for all three, which I'm not sure is true.

I was initially surprised at how asymmetric the > and < cases are, but upon reflection it makes sense: if you're going forward, it's easy to go the minimum number of steps, whereas if you're going backward, you need to compensate.

spec.bs
```diff
@@ -403,6 +423,139 @@ Each {{AppHistory}} object has an associated <dfn for="AppHistory">navigate meth
...
+
+    1. Let |targetEntry| be the [=session history entry=] in |navigableEntries| whose [=session history entry/app history key=] equals |key|. If no such entry exists, then:
+
+      1. [=Reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}}.
+
+      1. Abort these steps.
+
+      <p class="note">This can occur if the |appHistory| object's view of session history is outdated, which can happen for brief periods while all the relevant threads and processes are being synchronized in reaction to a history change (such as the user clearing their history).
+
+    1. Assert: |targetEntry| is not |navigable|'s [=navigable/active session history entry=].
```
domenic, line 532: Note that we synchronously returned a promise fulfilled with undefined if you do `appHistory.goTo(appHistory.current.key)`. Nevertheless, I'm not sure if this can be an assert or whether it needs to be checked again in this parallel queue.

My intuition is that an insert is safe because it seems impossible for a frame to get such an outdated view of its own active entry. And, we plan to abort any traversals if any other traversals start (not yet specced), so you couldn't do something like `const k = appHistory.entries()[appHistory.current.index - 1].key; appHistory.goTo(k); appHistory.goTo(k)` which would queue up two such traversals where the second one would hit this. But, maybe I'm not being devious enough.

domenic:
This has morphed into a pretty large overhaul of how state is tracked. I think it's almost done except the TODO to not fire navigatesuccess/navigateerror and resolve/reject the promise for canceled navigations. But in general it's much more robust.

spec.bs
```diff
@@ -346,6 +341,143 @@ Each {{AppHistory}} object has an associated <dfn for="AppHistory">navigate meth
...
+
+we need to ensure that when navigating to `url2`, we still have the {{Promise}} `p1` around so that we can reject it. We can't just get rid of any ongoing navigation promises the moment the second call to {{AppHistory/navigate()}} happens.
+
+We end up accomplishing all this using the following setup:
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing navigate event</dfn>, an {{AppHistoryNavigateEvent}} or null, initially null.
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">to-be-set serialized state</dfn>, a [=serialized state=] or null, initially null.
+
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">upcoming non-traverse navigation</dfn>, which is an [=app history API navigation=] or null, initially null.
```
natechapin, line 425: This is a better name than what I had proposed.

natechapin (approved):
Looks good!

Issue #117, opened 2021-05-20
Should event.navigationType for reloads be "reload" or "replace"?


domenic:
In working on #112 this came up. Currently it's specced as "replace", under the mental model that a reload is a same-URL replace.

Pros for treating reloads as "replace":

- Simpler mental model: replace/push/traverse are all very distinct and carve up the universe of navigations nicely.
- The distinction between `appHistory.navigate(appHistory.current.url, { replace: true })` and `appHistory.reload()`/`location.reload()` is not very important and developers probably want to treat them similarly.

Potential reasons to separate reloads into a new "reload" type:

- Maybe `navigate` handlers actually want to handle reloads specially, e.g. clearing UI state or something to give the user a "fresh" experience.
- Browsers treat reloads somewhat distinct from same-URL replaces in a couple cases and surfacing this might be good:
   - APIs like service workers and navigation timing
   - Some of the details around scroll position restoration seem to differ, although not very interoperably.

Yay295:
> APIs like service workers and navigation timing

Since the `PerformanceNavigation.type`/`PerformanceNavigationTiming.type` property has a distinct value for reloads, I think it would also make sense to distinguish reloads in this API. Or maybe even use that API in this API. The Navigation Timing API is still in development though so I don't know what exactly that would look like.

PR #118, opened 2021-05-20
Change appHistory.navigate(options) to appHistory.reload(options)

### LINKED ISSUE #112: We should add appHistory.reload(), and probably subsume appHistory.navigate(options)
As part of https://github.com/WICG/app-history/issues/67, @jakearchibald, @annevk, @smaug---- and I were thinking about how to allow developers to express reloads with app history, so that they could stop using `location`.

You could currently do `appHistory.navigate(appHistory.current.url)`, but this is not very explicit. Also, doing this fails to set the "reload-triggered navigation" flag which is visible to service workers, or set the navigationType visible to navigation timing.

So adding something like `appHistory.reload({ state, info })` makes sense.

But then, once we have that, we probably don't need `appHistory.navigate({ state, info })` anymore. Which is good, because that method was always a bit confusing.

domenic:
Closes #112. Note that the behavior differs in that previously appHistory.navigate(options) would do a same-URL replace, which is subtly different from a reload in observable ways.

Some follow-up discussion in #117.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/118.html" title="Last updated on May 20, 2021, 8:26 PM UTC (eb63794)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/118/e7f230f...eb63794.html" title="Last updated on May 20, 2021, 8:26 PM UTC (eb63794)">Diff</a>


README.md
```diff
@@ -636,6 +629,19 @@ await appHistory.navigate(url, { replace: true, state, navigateInfo });
...
+
+```js
+// Just like location.reload().
+await appHistory.reload();
+
+// Leave the state as-is, but pass some navigateInfo.
+await appHistory.reload({ navigateInfo });
+
+// Overwrite the state with a new value.
+await appHistory.reload({ state, navigateInfo });
```
natechapin, line 642: I don't instantly know how to implement this overwrite, but I assume it's solvable.

natechapin (approved):

Issue #124, opened 2021-05-27
First-class "client-side redirect" support


domenic:
The scenario shown in #121 uses `navigation.navigate(newURL, { replace: true })` in the middle of a `navigate` handler, to perform a "client-side redirect".

However, this is not optimal, as doing so makes the original navigation count as interrupted, and thus a "failure", with an `"AbortError"` `DOMException` as the failure reason. Notably it fires `navigateerror` on `navigation`, rejects `navigation.transition.finished`, and rejects the promise returned by `navigation.navigate()` if the navigation started that way.

There may be other downsides, e.g. the very-related discussion in #5 talks about the problems with losing same-siteness (for `Sec-Same-Site`) and form data. However, for purely client-side redirects I don't think those particular concerns apply. I guess maybe losing the `userInitiated` bit is in this vein, though.

We may want to contemplate a first-class redirect API, perhaps on the `NavigateEvent` object. The semantics would be something like a replace, but with the additional pieces:

- `navigation.transition.finished` delays settling until the new navigation settles
- `navigation.navigate()`'s return value delays settling until the new navigation settles
- `navigation.transition.rollback()` during the new navigation rolls back both navigations.
- `event.signal` for the original event does not fire `abort`.
- Maybe we carry over `event.userInitiated`
- Maybe we carry over `event.info` and `event.formData` unless told not to?
- There is still a period where `navigation.curentEntry` is the intermediate page (e.g. if you're navigating to `/b` and want to redirect to `/c`, `navigation.currentEntry` will briefly be on `/b` before eventually replacing with `/c`). There's no way of getting around that.

domenic:
I just realized something that ups the priority here.

If you're synchronously trying to do the redirect, as in this example:

```js
e.respondWith((async () => {
  if (shouldRedirect()) {
    await appHistory.navigate("/redirected"); // (*)
  } else {
    // ...
  }
})());
```

then you should *not* use `{ replace: true }`, since at the time the (*) line is running, the navigation hasn't happened yet: we're still in the synchronous phase of event dispatch (despite being inside an async function).

Whereas if you're asynchronously trying to do a redirect, as in this example 

```js
e.respondWith((async () => {
  if (await shouldRedirect()) {               // Note the await
    await appHistory.navigate("/redirected"); // (*)
  } else {
    // ...
  }
})());
```

then you *do* need to use `{ replace: true }`, because by that point the navigation has gone through and you're on the page that needs to get replaced.

[...]

PR #125, opened 2021-05-27
Introduce performance timeline API integration

### LINKED ISSUE #33: What design should be used for reporting navigation duration?

### LINKED ISSUE #59: currentchange is not suitable for measuring the time

domenic:
This replaces our previous broken idea of using `currentchange` for this. Closes #59. Closes #33. #14 remains to track whether `currentchange` is actually useful.

The main question is, what information should be included here, beyond the basics of URL/start time/duration? So far I added `success`, but I'm not sure whether that's useful. Other ideas include `navigationType`, `navigateInfo`, `userInitiated`, and `referrer` (or some other name for the previous URL).

See also https://github.com/WICG/app-history/issues/124 since I think interrupted navigations counting as failures would be pretty noisy to send to an analytics provider unless we also fix that.

@yoavweiss, I'd love your thoughts on this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,15 +222,6 @@ The current entry can be replaced with a new entry, with a new `AppHistoryEntry`
 
 - When using the `navigate` event to [convert a cross-document replace navigation into a same-document navigation](#navigation-monitoring-and-interception).
 
-In both cases, for same-document navigations a `currentchange` event will fire on `appHistory`:
-
-```js
-appHistory.addEventListener("currentchange", () => {
-  // appHistory.current has changed: either to a completely new entry (with new key),
-  // or it has been replaced (keeping the same key).
-});
-```
-
 ### Inspection of the app history list
 
 In addition to the current entry, the entire list of app history entries can be inspected, using `appHistory.entries()`, which returns an array of `AppHistoryEntry` instances. (Recall that all app history entries are same-origin contiguous entries for the current frame, so this is not a security issue.)
@@ -835,12 +826,6 @@ await appHistory.navigate("/1-b");
 
 This can be useful for cleaning up any information in secondary stores, such as `sessionStorage` or caches, when we're guaranteed to never reach those particular history entries again.
 
-### Current entry change monitoring
-
-The `window.appHistory` object has an event, `currentchange`, which allows the application to react to any updates to the `appHistory.current` property due to same-document navigations. This includes both navigations that change its value, and updates to its URL or state. This event cannot be intercepted or canceled, as it occurs after the navigation has already happened; it's just an after-the-fact notification.
-
-_TODO: Add examples of when this would be useful: [#14](https://github.com/WICG/app-history/issues/14). If no such examples can be found, delete this event._
-
 ### Performance timeline API integration
 
 The [performance timeline API](https://w3c.github.io/performance-timeline/) provides a generic framework for the browser to signal about interesting events, their durations, and their associated data via `PerformanceEntry` objects. For example, cross-document navigations are done with the [navigation timing API](https://w3c.github.io/navigation-timing/), which uses a subclass of `PerformanceEntry` called `PerformanceNavigationTiming`.
@@ -881,7 +866,6 @@ Between the per-`AppHistoryEntry` events and the `window.appHistory` events, as 
     1. `appHistory.current` fires `navigatefrom`.
     1. `location.href` updates.
     1. `appHistory.current` updates. `appHistory.transition` is created.
-    1. `currentchange` fires on `window.appHistory`.
     1. `appHistory.current` fires `navigateto`.
     1. Any now-unreachable `AppHistoryEntry` instances fire `dispose`.
     1. The URL bar updates.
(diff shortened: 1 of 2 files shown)


yoavweiss:
Apologies for the delay in reviewing this! Might be interesting to discuss this with the WG. 
One obvious question is whether we want a new entry type or if this can fit as a new [NavigationType](https://www.w3.org/TR/navigation-timing-2/#dom-navigationtype) on an [PerformanceNavigationTiming](https://www.w3.org/TR/navigation-timing-2/#dom-performancenavigationtiming) entry. 
That's at least the route we said we'd take with BFCache navigations, so might be good to coalesce all navigations on a similar API shape.

WDYT?

domenic:
> One obvious question is whether we want a new entry type or if this can fit as a new NavigationType on an PerformanceNavigationTiming entry.

Do you have thoughts on how that would work? In particular:

- Single-page app navigations can be single-page variants of each navigation type (and can be a replace as well, which is not captured by NavigationType). I.e. they are not a new type of navigation that sits alongside those.

- None of the fields on PerformanceNavigationTiming and PerformanceResourceTiming, except maybe `type`, make sense for single-page app navigations.

yoavweiss:
What we [talked](https://w3c.github.io/web-performance/meetings/2021/2021-01-21/index.html) about for BFCache is that it'd be interesting to [add](https://w3c.github.io/web-performance/meetings/2021/2021-02-18/index.html#h.vpaslvqqe1ab) a `NavigationID` to various PerformanceEntries, to be able to correlate them with the navigation that kicked them off. The current thinking should be summed up in this [design doc](https://docs.google.com/document/d/1LnHxd2RrY0vPI8ORYfO9HdnNs44zYmsX6_weV95BHZc/edit#heading=h.cbmzekd48lmj)



> * Single-page app navigations can be single-page variants of each navigation type (and can be a replace as well, which is not captured by NavigationType). I.e. they are not a new type of navigation that sits alongside those.

Looking at the list, I can see how they can be intermixed with "navigate" and "back_forward". That seems solvable (e.g. by adding 2 types for same-document navs)



> * None of the fields on PerformanceNavigationTiming and PerformanceResourceTiming, except maybe `type`, make sense for single-page app navigations.

The same is largely true for BFCache navigations. That's indeed unfortunate, but there were multiple voices among the API users that preferred the ergonomics and discoverability advantages of reusing the current "navigation" observers over having to kick off an observer for each new navigation type.  I'm happy to re-open that discussion if you disagree with that decision, but it'd be good to align both proposals on the same pattern.

domenic:
> Looking at the list, I can see how they can be intermixed with "navigate" and "back_forward". That seems solvable (e.g. by adding 2 types for same-document navs)

I don't quite understand the proposal. Are you saying we should expand `NavigationType` from `navigate`, `reload`, `back_forward`, `prerender` to also include `same-document-navigate`, `same-document-reload`, `same-document-replace`, and `same-document-back_forward`? That seems pretty awkward compared creating an orthogonal event...

> The same is largely true for BFCache navigations. That's indeed unfortunate, but there were multiple voices among the API users that preferred the ergonomics and discoverability advantages of reusing the current "navigation" observers over having to kick off an observer for each new navigation type. I'm happy to re-open that discussion if you disagree with that decision, but it'd be good to align both proposals on the same pattern.

I think it'd be good to reopen that discussion, especially since you could get the same benefits by using a new subclass but reusing the same `entryType` value.

But even beyond that, I would think that, given how same-document navigations are very different from cross-document ones, you'd want to use a separate `entryType` for those.

yoavweiss:
> `same-document-reload`, `same-document-replace`, and `same-document-back_forward`? That seems pretty awkward compared creating an orthogonal event...

I didn't really consider "reload" for same document navs and didn't consider "replace" to be a separate type. Agree that as you outlined it may be awkward, and it may be better to split that information to a separate attribute (e.g. a "same-document navigation" boolean). I'm not sure it justifies a completely separate entry though.



> I think it'd be good to reopen that discussion, especially since you could get the same benefits by using a fresh subclass of `PerformanceEntry` but reusing the same `entryType` value.

Can you expand on that?

Would you be interested in presenting to the WebPerfWG your proposal in one of the future calls?

domenic:
> I didn't really consider "reload" for same document navs and didn't consider "replace" to be a separate type.

Yeah, replace is a very separate type in specs and implementations, and we have some evidence developers want to handle them differently. So for single-page navs at least they should be separate.

> Can you expand on that?

What would you like me to expand on? The suggestion is to just have a subclass of `PerformanceEntry`, e.g. `BFCacheNavigationEntry`, which has `entryType === "navigation"`. Then all the `entryType`-based filtering/observing APIs would work, but you wouldn't have a violation of the Liskov Substitution Principle.

> Would you be interested in presenting to the WebPerfWG your proposal in one of the future calls?

Is there any way to have the discussion asynchronously?

domenic:
Based on some offline discussion with web developers, I got a few points of feedback:

- They reiterated that single-page navigations are very different from cross-document navigations, and they'd be surprised by getting a `PerformanceNavigationTiming` with a bunch of zero fields for something that doesn't apply. They have separate (usually application-specific) metrics they want to track for single-page navigations, and having a bunch of useless fields would make it confusing.
- To my point in the OP about "what information to include", they thought that just `success` was fine. The other things in the OP are kind of "facts about the navigation", which is best captured by listening for the `navigate` event. Whereas the performance timing integration should focus on performance and success/failure statistics.

I'll merge this for now, but happy to continue discussion, especially as we start implementing!

PR #129, opened 2021-06-15
Fire navigate events on all windows affected during traversal

### LINKED ISSUE #114: navigate events on traversal
In trying to spec `appHistory.back()`, I realized that there's a whole world of complexity around traversals that we seem to have forgotten, which is that history traversals can navigate multiple windows. Should all of them get `navigate` events? Which of them should be cancelable? Which should have `canRespond` true?

See also #78 and #32 which touch on some related questions, but no issue so far has tackled the multi-window issue I believe. There is currently a spec ("Modify the traverse the history by a delta algorithm" in https://wicg.github.io/app-history/#navigate-algorithm-patches) but it doesn't appear to acknowledge the possibility of multiple windows being traversed, so I need to rewrite it.

Recapping some guiding principles from previous discussions:

- If you get traversed and can either cancel it or `respondWith()` to replace it with a same-document nav, then you should get a `navigate` event. Conversely, if you can do neither of those things, then don't fire the `navigate` event, since it's not actionable.

- Same-document traversals always have `canRespond` true, and thus always fire `navigate` events. A SPA needs to be able to react to back-navigations just as much as it does to link clicks or other such things.

- Cross-document traversals must have `canRespond` false because converting those into same-document ones is just confusing.

- Developers are interested in canceling user-initiated traversals (#32) although currently we've specced this as disallowed.

I'll add my own additional principle, which hasn't been previously discussed: you shouldn't be able to cancel a traversal (using `event.preventDefault()`) unless "you" initiated it. I.e. if you're getting navigated because some other frame called `history.back()`, then the event should be uncancelable.

I think this leads to a model like the following:

1. Let currentWindow be the Window of the History or AppHistory object, or of the top-level browsing context being displayed while the user is pressing the back button.

1. Go to the browser process to figure out what windows a given traversal will affect, and whether the traversal will be same- or cross-document.

1. For each such Window w:

    1. Let canRespond be true if it's same-document, false otherwise.
    
    1. Let cancelable be true if this navigation is programatically-initiated and w = currentWindow; false otherwise. (We may also set it to true in some user-initiated cases per #32 in the future.)
    
    1. If canRespond is true or cancelable is true:
      
        1. Queue a task on w to fire the navigate event with cancelable and canRespond set appropriately.
        
        1. If the event is canceled, then bail out of the whole traversal (like when the user allows a beforeunload prompt to cancel the traversal), and reject the return value of `appHistory.back()` or whatever with an "AbortError" DOMException.
        
        1. If they called `respondWith()`, then...
          
            1. Track the promise given to respondWith. Maybe all the promises should be `Promise.all()`ed together to feed back into the return value of `appHistory.back()`? Or maybe only the one where w = currentWindow? I lean toward the latter.

            1. I don't think we need to do anything else? The history traversal was already underway, and it was already going to be same-document. So we just let it go through?

Ideally we'd have this be done in a single centralized algorithm, probably as patches to Jake's ["apply the history step"](https://whatpr.org/html/6315/history.html#apply-the-history-step). In this case we'd probably do these... after the beforeunload prompt portion of "apply the history step"?? (Step 13 currently.)

Then `history.back()` and `appHistory.back()` and user-initiated back button pressing all go through this centralized algorithm.

How does this sound, @jakearchibald @natechapin? I know in the short term in Chromium our implementation is firing the `navigate` event entirely within the renderer process, only for `appHistory.back()`, but as the vision for the ultimate architecture, does this make sense?

domenic:
Closes #114. @jakearchibald, can you review and also help with the TODO on figuring out the right step?

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/129.html" title="Last updated on Jun 16, 2021, 5:58 PM UTC (687f148)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/129/3794b90...687f148.html" title="Last updated on Jun 16, 2021, 5:58 PM UTC (687f148)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -47,6 +47,23 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: exceptionsEnabled; url: browsing-the-web.html#exceptions-enabled
   type: method
     for: Document; text: open(unused1, unused2); url: multipage/dynamic-markup-insertion.html#dom-document-open
+spec: html; urlPrefix: https://whatpr.org/html/6315/
+  type: dfn
+    text: traversable navigable; for: navigable; url: history.html#nav-traversable
+    text: current session history entry; for: navigable; url: history.html#nav-current-history-entry
+    text: active session history entry; for: navigable; url: history.html#nav-active-history-entry
+    text: get the session history entries; for: navigable; url: history.html#getting-session-history-entries
+    text: session history traversal queue; for: traversable navigable; url: history.html#tn-session-history-traversal-queue
+    text: current session history step; for: traversable navigable; url: history.html#tn-current-session-history-step
+    text: get all history steps; for: traversable navigable; url: history.html#getting-all-history-steps
+    text: step; for: session history entry; url: history.html#she-step
+    text: apply the history step; url: history.html#apply-the-history-step
+    text: checkForUserCancelation; for: apply the history step; url: history.html#apply-the-history-step
+    text: initiatorToCheck; for: apply the history step; url: history.html#apply-the-history-step
+    text: containing navigable; for: browsing context; url: browsers.html#bc-navigable
+    text: session; for: browsing context group; url: browsers.html#bcg-session
+    text: active document; for: navigable; url: history.html#nav-document
+    text: navigable; url: history.html#navigable
 spec: uuid; type: dfn; urlPrefix: https://wicg.github.io/uuid/
   text: generate a random UUID; url: #dfn-generate-a-random-uuid
 </pre>
@@ -612,10 +629,11 @@ The <dfn attribute for="AppHistoryDestination">sameDocument</dfn> getter steps a
 <h3 id="navigate-event-firing">Firing the event</h3>
 
 <div algorithm="fire a traversal navigate event">
-  To <dfn>fire a traversal `navigate` event</dfn> at an {{AppHistory}} |appHistory| given a [=session history entry=] <dfn for="fire a traversal navigate event">|destinationEntry|</dfn>, a boolean <dfn for="fire a traversal navigate event">|isSameDocument|</dfn>, an optional [=user navigation involvement=] <dfn for="fire a traversal navigate event">|userInvolvement|</dfn> (default "<code>[=user navigation involvement/none=]</code>"), and an optional JavaScript value |info| (default undefined):
+  To <dfn>fire a traversal `navigate` event</dfn> at an {{AppHistory}} |appHistory| given a [=session history entry=] <dfn for="fire a traversal navigate event">|destinationEntry|</dfn>, an optional [=user navigation involvement=] <dfn for="fire a traversal navigate event">|userInvolvement|</dfn> (default "<code>[=user navigation involvement/none=]</code>"), and an optional JavaScript value |info| (default undefined):
 
   1. Let |destinationURL| be |destinationEntry|'s [=session history entry/URL=].
   1. Let |destinationState| be |destinationEntry|'s [=session history entry/app history state=].
+  1. Let |isSameDocument| be true if |destinationEntry|'s [=session history entry/document=] is equal to |appHistory|'s [=relevant global object=]'s [=associated Document=]; otherwise false.
   1. Let |event| be the result of [=creating an event=] given {{AppHistoryNavigateEvent}}, in |appHistory|'s [=relevant Realm=].
   1. Set |event|'s [=AppHistoryNavigateEvent/destination entry=] to |destinationEntry|.
   1. Return the result of performing the [=inner navigate event firing algorithm=] given |appHistory|, |event|, "{{AppHistoryNavigationType/traverse}}", |isSameDocument|, |destinationURL|, |destinationState|, |userInvolvement|, |info|, and null.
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -612,10 +629,11 @@ The <dfn attribute for="AppHistoryDestination">sameDocument</dfn> getter steps a
...
+  To <dfn>fire a traversal `navigate` event</dfn> at an {{AppHistory}} |appHistory| given a [=session history entry=] <dfn for="fire a traversal navigate event">|destinationEntry|</dfn>, an optional [=user navigation involvement=] <dfn for="fire a traversal navigate event">|userInvolvement|</dfn> (default "<code>[=user navigation involvement/none=]</code>"), and an optional JavaScript value |info| (default undefined):
 
   1. Let |destinationURL| be |destinationEntry|'s [=session history entry/URL=].
   1. Let |destinationState| be |destinationEntry|'s [=session history entry/app history state=].
+  1. Let |isSameDocument| be true if |destinationEntry|'s [=session history entry/document=] is equal to |appHistory|'s [=relevant global object=]'s [=associated Document=]; otherwise false.
```
jakearchibald, line 636: Eventually it might be best to compare the "document state" between the active entry and the destination entry, as this caters for cases where the document is null.

Although, I guess conveniently, the document will never be null in this case.

spec.bs
```diff
@@ -989,6 +997,43 @@ With the above infrastructure in place, we can actually fire and handle the {{Ap
...
+  The <a spec="HTML">traverse the history by a delta</a> algorithm will be totally re-written as part of <a href="https://github.com/whatwg/html/pull/6315">the session history rewrite</a>. Here we reproduce the final version of the algorithm, after both that rewrite and with appropriate app history updates. Recall that per [[#user-initiated-patches]] we have introduced a |userInvolvement| argument, so the arguments are now |delta|, |source browsing context|, and |userInvolvement|.
+
+  1. Let |traversable| be |source browsing context|'s [=browsing context/containing navigable=]'s [=navigable/traversable navigable=].
+  1. Let |initiatorOrigin| be |source browsing context|'s [=active document=]'s [=Document/origin=].
+  1. [=parallel queue/enqueue steps|Enqueue the following steps=] to |traversable|'s [=traversable navigable/session history traversal queue=]:
+    1. Let |allSteps| be the result of [=traversable navigable/getting all history steps=] for |traversable|.
+    1. Let |targetStep| be TODO compute using [=traversable navigable/current session history step=], |allSteps|, and |delta|. Will be failure if there is no step |delta| away from the current one.
```
jakearchibald, line 1009: Something like:

```suggestion
    1. Let |currentStepIndex| be the index of the [=traversable navigable/current session history step=] within |allSteps|.
    1. Assert: |currentStepIndex| is not failure.
    1. Let |targetStepIndex| be |currentStepIndex| plus |delta|.
    1. If |allSteps|[|targetStepIndex|] does not [=set/exist=], then return.
    1. [=Apply the history step=] |allSteps|[|targetStepIndex|] to |traversable| with true, <var ignore>step</var>, |initiatorOrigin|, and |userInvolvement|.
```

Although there's no "index of". Maybe that should be in infra?

spec.bs
```diff
@@ -989,6 +997,43 @@ With the above infrastructure in place, we can actually fire and handle the {{Ap
...
+    1. If |targetStep| is failure, then return.
+    1. [=Apply the history step=] |targetStep| to |traversable| with true, <var ignore>step</var>, |initiatorOrigin|, and |userInvolvement|.
+</div>
+
+<div algorithm="apply the history step">
+  Modify the <a spec="HTML">apply the history step</a> algorithm as follows. Change <var ignore>checkForUserCancellation</var> to |fireBeforeunloadAndNavigate|. Add the |userInvolvement| parameter. Then, insert the following step after step 12 (which assembles |toTraverse| and other lists) but before step 13 (which checks if unloading is user-cancelled):
+
+  1. If |fireBeforeunloadAndNavigate| is true and the result of <a>firing traversal `navigate` events</a> given |toTraverse|, <var ignore>step</var>, <var ignore>initiatorToCheck</var>, and |userInvolvement| is false, then return.
```
jakearchibald, line 1017: Nit: It seems weird to me when an algorithm returns a boolean like this, but the name of the algorithm doesn't suggest what the return means. That's why I went for "checking if unloading is user-cancelled", but I guess the flip side of that is it isn't clear that the algorithm also fires events.

spec.bs
```diff
@@ -989,6 +997,43 @@ With the above infrastructure in place, we can actually fire and handle the {{Ap
...
+
+<div algorithm>
+  To <dfn>fire traversal `navigate` events</dfn> for a [=list=] of [=navigables=] |toTraverse|, an integer |step|, an [=origin=] |initiatorOrigin|, and a [=user navigation involvement=] |userInvolvement|:
+
+  1. Let |overallResult| be true.
+  1. Let |totalTasks| be the [=list/size=] of |toTraverse|.
+  1. Let |completedTasks| be 0.
+  1. [=list/For each=] |navigable| of |toTraverse|, [=queue a global task=] on the [=history traversal task source=] given |navigable|'s [=navigable/active document=]'s [=relevant global object=] to run these steps:
+      1. Let |destinationEntry| be the item in the result of [=navigable/getting the session history entries=] for |navigable| that has the greatest [=session history entry/step=] less than or equal to |step|.
+      1. If |destinationEntry|'s [=session history entry/document=] is not equal to |navigable|'s [=navigable/active document=], and |initiatorOrigin| is not [=same origin-domain=] with |navigable|'s [=navigable/active document=]'s [=Document/origin=], then abort these steps.
```
jakearchibald, line 1028: Again, this could compare document states rather than documents, but it doesn't really matter.

It would matter in cases where both documents could be null, which would be unintentionally equal.

jakearchibald (approved):

spec.bs
domenic: Yeah, I agree it's a bit iffy but I'm modeling it after the existing "fire an event" algorithm.

spec.bs
domenic: A navigable's active document is never null, right?

spec.bs
domenic: I think "index of" is relatively self-explanatory so I'm not too bothered about it not existing in Infra. Thanks, this is perfect!

domenic:
@natechapin, now I'd like your review. This doesn't yet account for the complexities of goTo() around passing info and dealing with the promise return value. However it shows the architecture that we'd use as a base. Roughly, any history traversal (whether via `history.go()` or using browser UI or, eventually, `appHistory.back()`):

1. Queues a task/IPC to the browser process asking for the traversal to happen
2. The browser process assembles a list of frames that are going to get navigated, and for each of them, maybe fires a navigate event. (Whether the event is fired/is cancelable/is canRespond true is computed at this time and subject to things we discussed previously. Most importantly, only same-origin-domain frames can cancel, and then only if it was programmatically initiated.)
3. If the event ends up canceled in any of the windows, then abort the history traversal---similar to the user clicking no on a beforeunload prompt.
4. Now proceed onward to doing beforeunload prompts in all unloaded frames. This might also abort the history traversal.
5. If the history traversal still goes through, then do the traversals, eventually posting a task/IPC into each renderer process to complete each one.

I realize this isn't how we currently implement things, and might take some work to get there. In particular, currently we don't yet support traversing multiple frames at once (due to "the bug"), and once we do, we probably need extra code to handle beforeunload in multiple frames, so I imagine firing navigate in multiple frames might also be difficult. But does it seem directionally reasonable?

@csreis might also be good to review the above flow, although I'm not sure if he wants to review the actual spec text.

natechapin (approved):
This is fine as far as I'm concerned, with the caveat that I'm still not sold on allowing `navigate` to cancel a cross-document history traversal.

PR #130, opened 2021-07-16
Fix signal specification

domenic:
Actually implementing this uncovered a few spec bugs. These edits largely mirror the structure in https://chromium-review.googlesource.com/c/chromium/src/+/2877951.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/130.html" title="Last updated on Jul 16, 2021, 8:53 PM UTC (7624852)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/130/72754f7...7624852.html" title="Last updated on Jul 16, 2021, 8:53 PM UTC (7624852)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -182,6 +182,8 @@ Each {{AppHistory}} object has an associated <dfn for="AppHistory">current index
 
 Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing navigate event</dfn>, an {{AppHistoryNavigateEvent}} or null, initially null.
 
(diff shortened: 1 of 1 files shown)


domenic:
Per offline discussion I think the spec is more broken than this makes it appear. E.g. the "navigate method call promise" isn't sticking around long enough for us to reject it on aborted navigations, which is bad.

It might be better to build this on top of #109 where the "navigate event side table" gives us a clear place to put the signal.

domenic:
Rolling into #109.

Issue #139, opened 2021-07-28
Should event.destination be an AppHistoryEntry for "traverse" navigations


domenic:
Spinning off from #97, re-summarizing based on the current state of things. Orthogonal to #138.

For "traverse" navigations, we currently have `event.destination` in the navigate event being a static snapshot saying where we're heading. But it could just as easily be the `AppHistoryEntry` we're heading to. The [`AppHistoryEntry` API](https://wicg.github.io/app-history/#apphistoryentry-class) is a superset of the [`AppHistoryDestination` API](https://wicg.github.io/app-history/#navigate-event-destination).

The differences (benefits?) would be:

- If it's an `AppHistoryEntry`, it will be "live". If you hold on to it, its `url` and `getState()` can update, and its `index` might become -1 if it gets kicked out of the app history list.
- If it's an `AppHistoryEntry`, you could attach event handlers for the [per-entry events](https://github.com/WICG/app-history#per-entry-events)

The drawbacks would be:

- It might be confusing that `event.destination` is live for traverse navigations but a static snapshot for push/replace/reload navigations.
- The reload case is especially tricky. We can't use `AppHistoryEntry` for reload navigations because given `appHistory.reload({ state: newState })`, we need to have `event.destination.getState()` return the new state, not the old state of the `AppHistoryEntry`. So although you might expect to get a real `AppHistoryEntry` for reloads, since like a traverse they're going to somewhere that already exists, you wouldn't get one.

Overall, making this change doesn't seem that compelling to me, but I'd like to keep this tracking issue open to see if people have other ideas.

bathos:
As someone doing kill-then-resurrect navigate handling (cancel initial attempt, perform transition, finally “commit” the original nav if the transition succeeds), I appreciate the snapshot behavior. It allows keeping the original navigate event around for that last step, where we’re trying (typically) to recreate the original nav attempt to the extent possible. If it weren’t a snapshot, we’d have to make the snapshot manually. That’s not hard, but it’s not super obvious that you’d need to, I think; it seems like an avoidable booby trap?

(I suspect there are cases beyond this specific pattern, too, where people are likely to be attracted to just keeping a navigation event around for “recreating” after something does or doesn’t happen — e.g. in “discard unsaved changes” flows.)

Not a strong opinion — just some feedback about why I think the snapshot behavior provides a bit of harder-to-screw-it-up-by-accident safety (and sometimes also convenience).

domenic:
Let's stick with the snapshot behavior!

[...]

PR #140, opened 2021-07-28
Create app_history.d.ts

tbondwilkinson:


app_history.d.ts
```diff
@@ -0,0 +1,99 @@
...
+interface AppHistory extends EventTarget {
+  readonly current: AppHistoryEntry|null;
+  readonly transition: AppHistoryTransition|null;
+
+  entries(): AppHistoryEntry[];
+
+  readonly canGoBack: boolean;
+  readonly canGoForward: boolean;
+
+  navigate(url: string, options?: AppHistoryNavigateOptions): Promise<void>;
```
frehner, line 14: I believe you need the overload for this for both options, right? 

```.ts
 navigate(options: AppHistoryFullOptions): Promise<void>;
 navigate(
    url?: string,
    options?: AppHistoryNavigateOptions
  ): Promise<void>;
  
 interface AppHistoryPushOrUpdateFullOptions extends AppHistoryNavigateOptions {
   url?: string;
 }
  ```

File: app_history.d.ts [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+
+  respondWith(newNavigationAction: Promise<void>): void;
+}
+
+interface AppHistoryNavigateEventInit extends EventInit {
+  navigationType?: AppHistoryNavigationType;
+  canRespond?: boolean;
+  userInitiated?: boolean;
+  hashChange?: boolean;
+  destination?: AppHistoryDestination;
```
frehner, line 84: I could be wrong, but is a destination just an `AppHistoryEntry`?

app_history.d.ts
frehner: Oh also, I believe that you can call `navigate()` with no options as well, correct?

domenic: The old no-url navigate() variant still exists in Chrome, but not in the spec; in the spec it was replaced by reload().

app_history.d.ts
domenic: No; see #139.

app_history.d.ts
```diff
@@ -0,0 +1,99 @@
...
+  readonly canGoForward: boolean;
+
+  navigate(url: string, options?: AppHistoryNavigateOptions): Promise<void>;
+  reload(options?: AppHistoryReloadOptions): Promise<void>;
+
+  goTo(key: string, options?: AppHistoryNavigationOptions): Promise<void>;
+  back(options?: AppHistoryNavigationOptions): Promise<void>;
+  forward(options?: AppHistoryNavigationOptions): Promise<void>;
+
+  onnavigate: ((ev: Event) => any)|null;
```
domenic, line 21: It looks like based on https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L2012 you'd want to set the `this` type here.

Also, you should use `AppHistoryNavigateEvent` for `onnavigate`, and `ErrorEvent` for `onnavigateerror`.

Finally, I'm a bit surprised by the `any` instead of `void` here, but that does seem to follow dom.d.ts conventions, so maybe it's correct. I did notice dom.d.ts uses `any` instead of `unknown` for `history.state` though, so maybe dom.d.ts is just a bit behind on best practices?

app_history.d.ts
domenic: Nit: destination is not optional in the IDL. (I.e. you cannot do `new AppHistoryNavigateEvent("x", {})`, you need to include `{ destination: somePreexistingDestination }`.)

Same for `signal`.

app_history.d.ts
```diff
@@ -0,0 +1,99 @@
...
+  hashChange?: boolean;
+  destination?: AppHistoryDestination;
+  signal?: AbortSignal;
+  formData?: FormData|null;
+  info?: unknown;
+}
+
+interface AppHistoryDestination {
+  readonly url: string;
+  readonly attribute: string;
```
domenic, line 92: Looks like a mistake

app_history.d.ts
```diff
@@ -0,0 +1,99 @@
...
+interface AppHistoryNavigateOptions extends AppHistoryNavigationOptions {
+  state?: unknown;
+  replace?: boolean;
+}
+
+interface AppHistoryReloadOptions extends AppHistoryNavigationOptions {
+  state?: unknown;
+}
+
+declare class AppHistoryNavigateEvent extends Event {
```
domenic, line 64: When do you choose `declare class` vs. `interface`? On the spec side everything is a class, but most of them have constructors that throw if you try to use them.

dom.d.ts seems to do something strange for most classes where they do e.g. `interface HTMLBaseElement { ... }` and then

```js
declare var HTMLBaseElement: {
    prototype: HTMLBaseElement;
    new(): HTMLBaseElement;
};
```

which doesn't really make sense to me...

/cc @saschanaz who I believe helps maintain dom.d.ts.

app_history.d.ts
bathos: I discovered by accidental earlier today that the impl in Canary does treat destination as nullable. Does that merit a Chromium issue or is it just a WIP artifact?

app_history.d.ts
saschanaz: That's [a decision from TypeScript team](https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/858#issuecomment-644294982) that [I don't exactly follow](https://github.com/microsoft/TypeScript/issues/39054#issuecomment-644299696), but one of the reasons was that the `interface` pattern is easier to extend than a class.

/cc @sandersn

app_history.d.ts
domenic: It looks like the issue is with both the spec and implementation in that they mark the init argument to the constructor as optional, when they should not. I'll fix the spec and file an issue.

app_history.d.ts
domenic: Wow, yeah, that's bizarre that they think the DOM classes are not classes. I guess that makes me lean toward using `declare class` for all of them?

What's the difference between just `interface` and the `interface` + `declare var` pattern?

saschanaz: Just `interface` does not allow reference to HTMLBaseElement e.g. `instanceof HTMLBaseElement` since it's not a value but a type. (See [demo](https://www.typescriptlang.org/play?ssl=11&ssc=28&pln=1&pc=1#code/JYOwLgpgTgZghgYwgAgGIHt0FEA2EC2E4yA3gL4CwAUNQBTnKgDOYcIS6MamuBRYASgDc1aqEixEKAEJwovQsXLUAJhAQ45KAG5zks+XkVgAXKWrJLyAA5R0YewE9rEMwYX8RVK8hAQA7rQCbnIe4F6UNFT0ZIwgLGwcXO5G-MJAA))

saschanaz: >I guess that makes me lean toward using declare class for all of them?

Yeah, it shouldn't harm here.

app_history.d.ts
tbondwilkinson: I think it probably has to do with covariant typing, so the spec is saying I explicitly never care about the return value of this function. I might be able to get away with "unknown" or "never" there instead. "void" would mean that no function is allowed to return a value.

I went with unknown for getState() because I wanted people to cast their states. "any" would probably be more TS-standard for existing APIs, but it seems weird to me to let anys leak in like that. But I'm not 100% what the TS philosophy is on these and why.

app_history.d.ts
tbondwilkinson: I'll go with declare class

app_history.d.ts
```diff
@@ -0,0 +1,98 @@
...
+  readonly canGoForward: boolean;
+
+  navigate(url: string, options?: AppHistoryNavigateOptions): Promise<void>;
+  reload(options?: AppHistoryReloadOptions): Promise<void>;
+
+  goTo(key: string, options?: AppHistoryNavigationOptions): Promise<void>;
+  back(options?: AppHistoryNavigationOptions): Promise<void>;
+  forward(options?: AppHistoryNavigationOptions): Promise<void>;
+
+  onnavigate: ((this: AppHistory, ev: Event) => any)|null;
```
domenic, line 21:
```suggestion
  onnavigate: ((this: AppHistory, ev: AppHistoryNavigateEvent) => any)|null;
```

app_history.d.ts
```diff
@@ -0,0 +1,98 @@
...
+  onnavigatesuccess: ((this: AppHistory, ev: Event) => any)|null;
+  onnavigateerror: ((this: AppHistory, ev: Event) => any)|null;
```
domenic, line 23:
```suggestion
  onnavigateerror: ((this: AppHistory, ev: ErrorEvent) => any)|null;
```

app_history.d.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  onnavigateto: ((this: AppHistoryEntry, ev: Event) => any)|null;
+  onnavigatefrom: ((this: AppHistoryEntry, ev: Event) => any)|null;
+  onfinish: ((this: AppHistoryEntry, ev: Event) => any)|null;
+  ondispose: ((this: AppHistoryEntry, ev: Event) => any)|null;
+}
+
+type AppHistoryNavigationType = 'reload'|'push'|'replace'|'traverse';
+
+declare class AppHistoryNavigationOptions {
```
domenic, line 51: So, the input dictionaries should stay as `interface`, sorry. This includes `AppHistoryNavigationOptions`, `AppHistoryNavigateOptions`, `AppHistoryReloadOptions`, and `AppHistoryNavigateEventInit`.

(Basically, Web IDL `interface` means TypeScript `class`, and Web IDL `dictionary` means TypeScript `interface`. I blame Web IDL for being confusing!)

domenic (changes requested):

app_history.d.ts
domenic: Yeah from what little reading I've done `unknown` seems really good for `state` and `info`. As for return types on event handlers, ¯\_(ツ)_/¯

app_history.d.ts
tbondwilkinson: Whoops got it.

app_history.d.ts
```diff
@@ -0,0 +1,98 @@
...
+  readonly hashChange: boolean;
+  readonly destination: AppHistoryDestination;
+  readonly signal: AbortSignal;
+  readonly formData: FormData|null;
+  readonly info: unknown;
+
+  respondWith(newNavigationAction: Promise<void>): void;
+}
+
+declare class AppHistoryNavigateEventInit extends EventInit {
```
domenic, line 79:
```suggestion
interface AppHistoryNavigateEventInit extends EventInit {
```

domenic (commented):
One last interface then we're good.

app_history.d.ts
tbondwilkinson: done!

domenic (approved):

app_history.d.ts
sandersn: Briefly, if a class needs to be extensible via merging (declaring twice), you can't use `declare class` because it's block-scoped and doesn't allow multiple declarations. You have to use the weird structure like for HTMLBaseElement.

In other words, this doesn't work:

```ts
// @filename: app_history.d.ts
declare class AppHistoryNavigateEvent extends Event {
  // properties from this PR
}

// @filename: somebody_else_extends_app_history.d.ts
declare class AppHistoryNavigateEvent extends Event {
  // properties that somebody else has added later
}
```

sandersn: If users are supposed to be able to tack properties onto instances of AppHistoryNavigateEvent, these should be written using the decomposed class pattern so that other d.ts files can declare those properties.

saschanaz: I don't think this expects any extension. And classes are still extendable by interfaces, no?

sandersn: > And classes are still extendable by interfaces, no?

Yes. The example code I gave is trying to provide typing for people adding expando properties to `AppHistoryNavigateEvent` instances, not people using `X extends AppHistoryNavigateEvent`.

saschanaz: Not sure I follow, I mean [declaration merging](https://www.typescriptlang.org/play?#code/MYGwhgzhAECCAO8ASBLCAXA9gJwJ4DkwA3FAczHQFMBRIygO3WgG8BfAbgFgAoHlRytgBmYYJTiJUGHAWJkKNOoxY9oa6JQAe8MPQAmmAArZM8QelwAuaPQCuAWwBGgrt1Y8e9SgHcJyNFh4hCTkVLQM6AAUAJQAdFo6+kYmZtgW7EA).

sandersn: Yep, that's what I'm talking about too. Declaration merging allows types to represent expando properties on values.

saschanaz: Yeah and that doesn't require decomposed class pattern, unless someone wants to add static properties (which doesn't work anyway in decomposed class pattern, unless there is no named interface for the constructor part)?

sandersn: Representing expando properties does require a decomposed class pattern + declaration merging as far as I know. Is there another way?

saschanaz: The following works as I said in https://github.com/WICG/app-history/pull/140#discussion_r679560568...

```ts
class AppHistoryNavigateEvent {};

interface AppHistoryNavigateEvent {
    expandoProperty: number;
}

new AppHistoryNavigateEvent().expandoProperty;
```

sandersn: Sorry, I didn't follow the link and assumed you meant merging two interfaces with the same name instead of merging a class and an interface.

That pattern is much easier to read, and the only thing it misses is the ability to polyfill by copying the entire declaration.

PR #142, opened 2021-08-02
add event map to AppHistory, for types of Event

samthor:
This will give types when writing code like:

```js
window.appHistory.addEventListener('navigate', (event) => {
  // "event" is known by TS to be of type AppHistoryNavigateEvent
});
```

### DIFF
--- a/app_history.d.ts
+++ b/app_history.d.ts
@@ -4,6 +4,7 @@ interface Window {
 
 interface AppHistoryEventMap {
   "navigate": AppHistoryNavigateEvent;
+  "navigatesuccess": Event;
   "navigateerror": ErrorEvent;
 }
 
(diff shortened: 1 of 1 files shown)


app_history.d.ts
```diff
@@ -2,6 +2,11 @@ interface Window {
   readonly appHistory: AppHistory;
 }
 
+interface AppHistoryEventMap {
+  "navigate": AppHistoryNavigateEvent;
+  "navigateerror": ErrorEvent;
```
domenic, line 7: Should we also add `navigatesuccess`? That's just `Event`, but it looks like that's still reasonably common: e.g. https://github.com/microsoft/TypeScript/blob/9cc30709380da8b212487f63d134b7be9b001511/lib/lib.dom.d.ts#L1916-L1931

domenic (commented):
Should we do the same thing for AppHistoryEntry? The events aren't implemented (or specced) yet but it seems like we're tracking the explainer more than the implementation or spec.

samthor:
Yeah, whether to include `Event` was a bit unclear but I see the precedent and it makes it clearer what to update if the `Event` types change. I'll update the PR.

app_history.d.ts
samthor: Done

domenic (approved):

PR #143, opened 2021-08-04
Handle navigation and traversal cancelation more uniformly

### LINKED ISSUE #141: Early bail out spec for canRespond = false, cancelable = false seems wrong
https://wicg.github.io/app-history/#navigate-event-firing step 8 will bail out if canRespond is false and cancelable is false. As part of that bailout, it will clean up the ongoing navigation, and resolve the promise.

However, canRespond = false and cancelable = false only happens for cross-document browser UI navigations. In those cases, as with all cross-document navigations, our current plan is to leave the promise unsettled. That might change in #95, but for now the spec seems broken.

domenic:
Previously, cancelation of queued-up traversals was ignored. Now it triggers the appropriate finalization algorithm.

The introduction of a "fired navigate event" boolean on the ongoing app history API navigation struct also ensures that we don't fire navigateerror on cases where we never fire the navigate event.

Also closes #141 while we're there.

Finally, this removes the microtask delay in "finalize with an aborted navigation error". This works poorly in practice, because if a navigation interrupts another navigation, you want the navigateerror for the old navigation to fire before the navigate for the new navigation.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/143.html" title="Last updated on Aug 4, 2021, 10:59 PM UTC (1dc470a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/143/1a837bf...1dc470a.html" title="Last updated on Aug 4, 2021, 10:59 PM UTC (1dc470a)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -416,6 +416,8 @@ const p2 = appHistory.navigate(url2);
 
 we need to ensure that when navigating to `url2`, we still have the {{Promise}} `p1` around so that we can reject it. We can't just get rid of any ongoing navigation promises the moment the second call to {{AppHistory/navigate()}} happens.
 
+We also need to ensure that, if we start a new navigation, navigations which have gotten as far as firing {{AppHistory/navigate}} events, but not yet as far as firing {{AppHistory/navigatesuccess}} or {{AppHistory/navigateerror}}, get [=finalized with an aborted navigation error=].
+
 We end up accomplishing all this using the following setup:
 
 Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing navigate event</dfn>, an {{AppHistoryNavigateEvent}} or null, initially null.
@@ -434,6 +436,7 @@ An <dfn>app history API navigation</dfn> is a [=struct=] with the following [=st
 
 * An <dfn for="app history API navigation">info</dfn>, a JavaScript value
 * A <dfn for="app history API navigation">returned promise</dfn>, a {{Promise}}
+* A <dfn for="app history API navigation">fired `navigate` event</dfn>, a boolean
 * A <dfn for="app history API navigation">cleanup step</dfn>, an algorithm step
 
 <p class="note">We need to store the {{AbortSignal}} separately from the [=app history API navigation=] struct, since it needs to be tracked even for navigations that are not via the app history APIs. So, we store it some of the time in the [=AppHistory/ongoing navigate event=]'s {{AppHistoryNavigateEvent/signal}} property, and the rest of the time in the [=AppHistory/post-navigate event ongoing navigation signal=].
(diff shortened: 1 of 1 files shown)


domenic:
Some analysis of whether this "finalizes with an aborted navigation error" often enough. In particular we need to think about the cases listed in https://github.com/whatwg/html/issues/6927 when the SDT or SDN is extended by `respondWith()`.

Example:

```js
  e.respondWith(new Promise(r => setTimeout(r, 1_000)));
});

await appHistory.navigate("#foo");

// What does this do?
const p1 = appHistory.navigate("#bar");
history.back();
```

This is a SDN-SDT case, which in the pre-app history world is straightforward: the SDN completes synchronously, and so the SDT then proceeds. (We would end up on #foo.)

In the app history world however, at the time `history.back()` is called, the navigation to `#bar` is still "ongoing". That is, `location.href` is updated to `#bar` already, but the loading spinner is still spinning, and `navigatesuccess` hasn't fired, and `p1` hasn't settled.

So we have to decide what it means to traverse back while a navigation is still ongoing. IMO the simplest is that the navigation counts as aborted: it didn't reach the end state, as signaled by the developer's promise passed to `respondWith()`.

Indeed, in this case we match the behavior for cross-document navs/traversals: per https://github.com/whatwg/html/issues/6927, XDN + XDT results in the XDN being canceled. So in this case developers could have  the intuition "using `respondWith()` to extend a same-document nav/traversal makes their overlap behave like the corresponding cross-document case".

The app history spec draft, after this PR, gives the desired result. Because of the clause saying "whenever you would normally cancel non-matured cross-document navigations, also inform app history about canceling navigation", we properly reject `p1` and fire `navigateerror`.

---

Let's consider the reverse case:

```js
  e.respondWith(new Promise(r => setTimeout(r, 1_000)));
});

await appHistory.navigate("#foo");

// What does this do?
const p1 = appHistory.back();
const p2 = appHistory.navigate("#bar");
```

This is a SDT-SDN case, which does not have interoperable behavior: Firefox queues up the SDT to happen after the SDN, whereas Chrome ignores the SDN. It's not clear what direction the spec will go in yet, but I would say that if the SDN is ignored, then `p2` should reject (and no `navigate` event should fire for it).

If you look at the corresponding XDT-XDN case, trying to port over our intuition from above, you find from https://github.com/whatwg/html/issues/6927 that the XDN is ignored in both Chrome and Firefox. So that indicates maybe Chrome's behavior is a bit more uniform.

The app history spec draft, after this PR, does *not* reject `p2` in such a situation. App history is only informed when the browser cancels navigations, not when it ignores them. Hmm.

natechapin (approved):

PR #146, opened 2021-08-10
Add appHistory.updateCurrent() to close #115

### LINKED ISSUE #115: Some non-navigating setState() is probably needed
The API currently only allows you to update the app history state via navigation. E.g., to get the equivalent of `history.pushState(newState, "", url)` you would do `appHistory.navigate(url, { state: newState })` and then intercept this using the `navigate` event to convert it into a single-page app navigation. The thinking here is to flip the traditional model on its head: instead of updating the document and then syncing the URL + history stack + state with a disconnected `history.pushState()` call, you initiate a navigation using `appHistory.navigate()`, and in a centralized fashion do SPA processing. [This explainer section](https://github.com/WICG/app-history#using-navigate-handlers) shows some of the advantages.

However, there are some cases where this model doesn't seem to work very well. That's when you need to update the current state in response to an external event, often caused by user interaction.

Here is an example. Consider a page with expandable/collapsable `<details>` elements. You want to store the expanded/collapsed state of these `<details>` elements in your app history state, so that when the user traverses back and forward through history, or restarts their browser, your app can read the restored app history state and expand the `<details>` elements appropriately, showing the user what they saw previously.

To do this using `history.replaceState()` is relatively easy (modulo the fact that `history.state` is fragile and terrible): you listen for `toggle` events and then do `history.replaceState()` to sync `history.state` with the `<details>`.

With app history it's more awkward. You could listen for the `toggle` event and then do `appHistory.navigate({ state: newState })` (or `appHistory.reload({ state: newState })` per #112). But then you need to have your `navigate` handler do nothing, because _the `<details>` element is already open_. This is certainly doable, but it points out how the model we're forcing you into is a bit awkward.

To summarize, not every state update *causes* a page update; instead, sometimes a state update is to *capture* an externally-driven page update.

For cases like these, we might need something like `appHistory.current.setState()` (briefly introduced, but removed in https://github.com/WICG/app-history/pull/61.)

Points of discussion:

- Is this a big problem worth solving? Or is it OK for cases like this to involve `navigate` handlers that purposefully do nothing?

- Is this true for just state, or are there cases where it makes sense for URLs too? You could imagine a site that wanted to store UI state like `<details>` open-ness in the URL, but I'm not sure we should design the API around such cases.

- If we add `setState()`, should it be able to set the state of any entry, or just the current one?

- If we add `setState()`, how do we encourage people to avoid it in cases where the state update should drive a page update (for which our existing pattern works best), and instead only use it in cases where a page update drives a state update? Just documentation?

- The above example shows page-driven state _updates_. Are there page-driven _history entry addition_ cases, where something like `history.pushState()` would make sense? (I hope not...)

domenic:
Review appreciated from any watchers on the explainer updates and d.ts updates.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 10, 2021, 5:32 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FWICG%2Fapp-history%2Fa57b672385a810b19d2b22d5c3f4307491e8bf20%2Fspec.bs&md-warning=not%20ready)

```
Error running preprocessor, returned code: 1.
Traceback (most recent call last):
  File "/sites/api.csswg.org/bikeshed/bikeshed.py", line 6, in 
    cli.main()
  File "/sites/api.csswg.org/bikeshed/bikeshed/cli.py", line 466, in main
    handleSpec(options, extras)
  File "/sites/api.csswg.org/bikeshed/bikeshed/cli.py", line 515, in handleSpec
    lineNumbers=options.lineNumbers,
  File "/sites/api.csswg.org/bikeshed/bikeshed/Spec.py", line 61, in __init__
    self.inputSource = InputSource(inputFilename, chroot=constants.chroot)
  File "/sites/api.csswg.org/bikeshed/bikeshed/InputSource.py", line 51, in __new__
    return UrlInputSource(sourceName, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'chroot'
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/app-history%23146.)._
</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,8 +82,8 @@ for (const entry of performance.getEntriesByType("same-document-navigation")) {
   - [New navigation API](#new-navigation-api)
     - [Example: using `info`](#example-using-info)
     - [Example: next/previous buttons](#example-nextprevious-buttons)
+  - [Setting the current entry's state without navigating](#setting-the-current-entrys-state-without-navigating)
   - [Per-entry events](#per-entry-events)
-  - [Current entry change monitoring](#current-entry-change-monitoring)
   - [Performance timeline API integration](#performance-timeline-api-integration)
   - [Complete event sequence](#complete-event-sequence)
 - [Guide for migrating from the existing history API](#guide-for-migrating-from-the-existing-history-api)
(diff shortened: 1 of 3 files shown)


natechapin (approved):

samthor:
Does this have any odd interactions with a currently occuring navigation? `getState()` appears on the event (as part of its destination) and the entry.

I presume since fetching the data is synchronous, it's possible (imagine user code which does this) to "get" the state earlier, update it, and still have the previous/old copy of the state—all while a navigation is occuring.

domenic:
> Does this have any odd interactions with a currently occuring navigation? getState() appears on the event (as part of its destination) and the entry.

They are disconnected. `appHistory.destination.getState()` returns a snapshot of the state passed to `appHistory.navigate(, { state })`, or the state stored in the existing history entry if that is left undefined and a traversal is being performed. `appHistory.updateCurrent()` updates the state returned by `appHistory.current.getState()`.

> I presume since fetching the data is synchronous, it's possible (imagine user code which does this) to "get" the state earlier, update it, and still have the previous/old copy of the state—all while a navigation is occuring.

I'm not sure I understand the scenario you're talking about. Could you give the example in code?

PR #156, opened 2021-08-18
Consolidate ongoing navigation signal tracking

domenic:
Follows https://chromium-review.googlesource.com/c/chromium/src/+/3102875 and https://chromium-review.googlesource.com/c/chromium/src/+/3104346, although the spec's "inform app history about canceling navigation" / "inform app history about canceling traversals" don't quite map to the implementation.

By introducing a guard in those algorithms that checks for the ongoing navigation signal, this fixes a spec bug where we would sometimes fire navigateerror even if there's no ongoing navigation (e.g. if window.stop() is called).

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/156.html" title="Last updated on Aug 18, 2021, 11:22 PM UTC (486d069)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/156/f647fa4...486d069.html" title="Last updated on Aug 18, 2021, 11:22 PM UTC (486d069)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -45,6 +45,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: historyHandling; url: browsing-the-web.html#navigation-hh
       text: navigationType; url: browsing-the-web.html#navigation-navigationtype
       text: exceptionsEnabled; url: browsing-the-web.html#exceptions-enabled
+    for: browsing context
+      text: discard; url: window-object.html#a-browsing-context-is-discarded
   type: method
     for: Document; text: open(unused1, unused2); url: multipage/dynamic-markup-insertion.html#dom-document-open
 spec: html; urlPrefix: https://whatpr.org/html/6315/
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1211,6 +1209,7 @@ The <dfn attribute for="AppHistoryDestination">sameDocument</dfn> getter steps a
   To <dfn>inform app history about canceling traversals</dfn> in a [=browsing context=] |bc|:
 
   1. Let |appHistory| be |bc|'s [=browsing context/active window=]'s [=Window/app history=].
+  1. If |appHistory|'s [=AppHistory/ongoing navigation signal=] is null, then return.
```
natechapin, line 1212: I'm not sure about this. We can have more than one ongoing traversal in flight (and at least in the chromium implementation, ongoing navigation signal might be null while we have one ore more ongoing traversals).

Two potential ideas:
* Assert that ongoing navigation siganl is null (i.e., we recently did the "inform app history about canceling navigation" steps), 
* Invoke "inform app history about canceling navigation" if ongoing navigation signal is non-null here to ensure it is handled before iterating over the ongoing traversals.

domenic: Good catch. After working through a variety of options here I ended up doing something similar to your second idea, although I went further and embraced the implementation even more. Let me know what you think, and note the new text in the "Canceling navigation and traversals" section.

natechapin (approved):

PR #157, opened 2021-08-19
Specify beforedispose on AppHistoryEntry

domenic:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/157.html" title="Last updated on Sep 1, 2021, 10:06 PM UTC (ea63fab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/157/1e5ae85...ea63fab.html" title="Last updated on Sep 1, 2021, 10:06 PM UTC (ea63fab)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -865,19 +865,19 @@ async function showPhoto(photoId) {
 
 Note how in the event handler for these events, `appHistory.current` will be set as expected (and equal to `e.target`), so that the event handler can use its properties and methods (like `id`, `url`, or `getState()`) as needed.
 
-Finally, there's a `dispose` event, which occurs when an app history entry is permanently evicted and unreachable: for example, in the following scenario.
+Finally, there's a `beforedispose` event, which occurs when an app history entry is about to be permanently evicted and unreachable: for example, in the following scenario.
 
 ```js
 const startingKey = appHistory.current.key;
 
 await appHistory.navigate("/1");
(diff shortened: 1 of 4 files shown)


natechapin (approved):

natechapin (approved):
This is fine for now, but we should make sure we really need to fire dispose events during unload. That could end up being a sharp edge.

domenic:
Given how much iteration we needed to go through on beforedispose implementation and tests, I've lost confidence in my ability to spec per-entry events without doing the implementation/tests in parallel. So I pared this PR down to only being about beforedispose (with some stub definitions for the others).

This should now match the implementation at https://chromium-review.googlesource.com/c/chromium/src/+/3107610, but note that hasn't gotten LGTM yet, so we'll see.

domenic:
After talking with some developers offline, it became clear that we didn't really need beforedispose for cross-document cases (i.e. the unload integration). It would, as @natechapin feared, mostly cause unnecessary work to be done. That is, most of the things you'd clean up in a `beforedispose` handler are things that would go away anyway when the document unloads.

For the cases where they really need to clean up something persistent, the existing `pagehide` event is the best route.

So, let me close this PR and open a new one with just the same-document case. And maybe it'll go back to `dispose` in that case; still thinking...

PR #158, opened 2021-08-20
Further improvements to ongoing navigation tracking

domenic:
Follows https://chromium-review.googlesource.com/c/chromium/src/+/3108511, and likely fixes a similar bug in the spec as that does in the implementation.

One notable change here is that we no longer have the "cleanup step" concept, since API navigations move around enough that we don't know at creation time the right mechanism for cleaning them up. Instead we follow the implementation, and have API navigations track their key, and a centralized "clean up an API navigation" algorithm first check if they're ongoing, then clean them up based on the key if not.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/158.html" title="Last updated on Aug 20, 2021, 6:02 PM UTC (a100524)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/158/33af84b...a100524.html" title="Last updated on Aug 20, 2021, 6:02 PM UTC (a100524)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -540,18 +540,18 @@ Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing navig
 
 Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing navigation signal</dfn>, which is an {{AbortSignal}} or null, initially null.
 
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing navigation</dfn>, which is an [=app history API navigation=] or null, initially null.
+
 Each {{AppHistory}} object has an associated <dfn for="AppHistory">upcoming non-traverse navigation</dfn>, which is an [=app history API navigation=] or null, initially null.
 
-Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing non-traverse navigation</dfn>, which is an [=app history API navigation=] or null, initially null.
-
-Each {{AppHistory}} object has an associated <dfn for="AppHistory">ongoing traverse navigations</dfn>, which is a [=map=] from strings to [=app history API navigations=], initially empty.
+Each {{AppHistory}} object has an associated <dfn for="AppHistory">upcoming traverse navigations</dfn>, which is a [=map=] from strings to [=app history API navigations=], initially empty.
 
 An <dfn>app history API navigation</dfn> is a [=struct=] with the following [=struct/items=]:
 
+* A <dfn for="app history API navigation">key</dfn>, a string or null
 * An <dfn for="app history API navigation">info</dfn>, a JavaScript value
 * An <dfn for="app history API navigation">serialized state</dfn>, a [=serialized state=] or null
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -572,27 +570,51 @@ An <dfn>app history API navigation</dfn> is a [=struct=] with the following [=st
...
+</div>
+
+<div algorithm>
+  To <dfn for="AppHistory">clean up an API navigation</dfn> given an {{AppHistory}} |appHistory| and an [=app history API navigation=] |navigation|:
+
+  1. If |appHistory|'s [=AppHistory/ongoing navigation=] is |navigation|, then set |appHistory|'s [=AppHistory/ongoing navigation=] to null.
+
+  1. Otherwise,
+
+    1. Assert: |navigation|'s [=app history API navigation/key=] is not null.
```
domenic, line 613: This assert is stronger than the implementation. Let me know if it's incorrect.

spec.bs
```diff
@@ -1109,10 +1131,8 @@ The <dfn attribute for="AppHistoryDestination">sameDocument</dfn> getter steps a
 <div algorithm>
   The <dfn>inner `navigate` event firing algorithm</dfn> is the following steps, given an {{AppHistory}} |appHistory|, an {{AppHistoryNavigationType}} |navigationType|, an {{AppHistoryNavigateEvent}} |event|, an {{AppHistoryDestination}} |destination|, a [=user navigation involvement=] |userInvolvement|, and a [=list=] of {{FormData}} [=FormData/entries=] or null |formDataEntryList|:
 
-  1. [=AppHistory/Promote the upcoming non-traverse navigation to ongoing=] given |appHistory|.
-  1. Let |ongoingNavigation| be null.
-  1. If |destination|'s [=AppHistoryDestination/key=] is null, then set |ongoingNavigation| to |appHistory|'s [=AppHistory/ongoing non-traverse navigation=].
-  1. Otherwise, if |appHistory|'s [=AppHistory/ongoing traverse navigations=][|destination|'s [=AppHistoryDestination/key=]] [=map/exists=], then set |ongoingNavigation| to |appHistory|'s [=AppHistory/ongoing traverse navigations=][|destination|'s [=AppHistoryDestination/key=]].
+  1. [=AppHistory/Promote the upcoming navigation to ongoing=] given |appHistory| and |destination|'s [=AppHistoryDestination/key=].
+  1. Let |ongoingNavigation| be |appHistory|'s [=AppHistory/ongoing navigation=].
```
domenic, line 1135: I factored out all these steps into a single helper algorithm, versus the implementation which spreads them out a bit. I am pretty sure they are equivalent (and maybe we could clean up the implementation), but it'd be good to double-check.

spec.bs
natechapin: I think you're right.

spec.bs
natechapin: Agreed, I'll switch to this.

natechapin (approved):

PR #164, opened 2021-09-07
Change the return value of navigating methods

### LINKED ISSUE #163: Providing a better per-navigation signal of when the new entry is created
In offline discussions with @tbondwilkinson, he noted that it's particularly difficult to get a handle to the entry you're navigating to, so that you can do things like setup per-entry event handlers or use its `key`. In particular the desire is something like

```js
appHistory.navigate(...);
// Now set up event handlers for the navigated-to entry
```

or

```js
appHistory.back();
// Now send the navigated-to key to some other system
```

<details>
<summary>Deeper analysis of what doesn't work very well</summary>

#148 can help a bit, but not entirely, because it's not clear which navigation caused the current entry change. In some cases that doesn't matter, but for some flows it does. So you'd need some impressive synchronization between the `navigate()`/`reload()`/`back()`/`forward()`/`goTo()` call and the `currentchange` handler. It might not even be possible to write such synchronization code in tricky cases, like if someone navigates inside of `navigate`/`beforedispose`/`navigatefrom`/`navigateto` handler.

You can kind of hack this in most cases using the following facts:

- After a same-document `navigate()` or `reload()` returns, but before the promise settles, `appHistory.current` will be updated. So code like

    ```js
    const p = appHistory.navigate(...);
    const navigatedTo = appHistory.current;
    await p;
    ```
    
    will get the correct value, usually.
    
    However, this can break if you do navigations inside of the `navigate`/`beforedispose`/`navigatefrom`/`navigateto` handlers.

- For a same-document traversal, you can know the entry ahead of time, e.g. with code like

  ```js
  const navigatedTo = appHistory.entries()[appHistory.current.index - 1];
  await appHistory.back(...);
  ```
  
  However, unlike same-document navigations, same-document traversals are async, and there's no signal as to when the entry actually changes.

An early idea we had was to just have the return value promises for `navigate()`/etc. fulfill with the destination entry. The problem with this is twofold:

- The promise doesn't settle until the full transition finishes. This might be bad for some use cases.

- If the transition fails, then the promise will reject, and you won't have the fulfillment value (i.e. the entry you care about). This is problematic because failed transitions don't actually prevent the entry from updating.
</details>

There are essentially three ways to solve this that I could think of:

1. A new object, e.g. `appHistory.destination`, which you would consult immediately after calling `navigate()`/`reload()`/`back()`/`forward()`/`goTo()`.

2. Change the semantics of `navigate()`/etc.'s return value. Instead of returning a promise for the full transition, return something like a promise for the destination entry that fulfills the moment the current entry changes, or some sort of container object, or...

3. Use a custom error subclass for the rejection reason of `navigate()`/etc.'s return value, which contains the destination entry. (Note, unlike (1) or (2), with (3) you do not get the handle to the entry ASAP, but instead only after the transition settles.)

Of these I think (2) is much better. So the question is, what should the return value be.

Here's one proposal: a plain object with two promise properties, e.g. `{ committed, finished }`.

- `finished` is a promise with the same semantics as the current return value.
- `committed` (unsure on the name) is a promise which fulfills the moment the URL/entry change goes through for the navigation, and fulfills with the new entry.

Both promises will still reject in cases where the navigation doesn't work at all, e.g. trying to navigate during beforeunload, or in an inactive document. (Or maybe the method should throw, now that that's an option?? Hmm.) But `committed` will generally fulfill much more often and more quickly.

I like this solution because I suspect that some call sites probably want `committed` semantics anyway, and don't need to know about the transition. And, it forces you to be very explicit about what you're `await`ing:

```js
await appHistory.navigate(...).committed; // wait for the entry change

await appHistory.navigate(...).finished;  // wait for the full transition
```

A potential extension of this solution is to add a `then()` method to the return value, which proxies to one of the promises. But, we've so far avoided creating non-promise thenables on the web platform, so I think I'm against this.

domenic:
Closes #163.

Given how previous attempts at speccing without implementing has gone, I'll delay merging this until I have a prototype ready... but people should feel free to review the explainer and .d.ts updates.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/164.html" title="Last updated on Sep 16, 2021, 6:14 PM UTC (189a7f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/164/dfcac27...189a7f0.html" title="Last updated on Sep 16, 2021, 6:14 PM UTC (189a7f0)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ function renderHomepage() {
 
   document.querySelector("#home-button").addEventListener("click", async e => {
     try {
-      await appHistory.goTo(homepageKey);
+      await appHistory.goTo(homepageKey).finished;
     } catch {
       // Fall back to a normal "forward" navigation
       appHistory.navigate("/");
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -256,18 +256,20 @@ function renderHomepage() {
...
 
-All of these methods return promises, because navigations can be intercepted and made asynchronous by the `navigate` event handlers that we're about to describe in the next section. There are then several possible outcomes:
+All of these methods return `{ committed, finished }` pairs, where both values are promises. This because navigations can be intercepted and made asynchronous by the `navigate` event handlers that we're about to describe in the next section. There are then several possible outcomes:
 
-- The `navigate` event responds to the navigation using `event.transitionWhile()`, in which case the promise fulfills or rejects according to the promise(s) passed to `transitionWhile()`. (However, even if the promise rejects, `location.href` and `appHistory.current` will change.)
+- The `navigate` event cancels the navigation without responding to it, in which case both promises reject with an `"AbortError"` `DOMException`, and `location.href` and `appHistory.current` stay on their original value.
```
posva, line 261: By _cancels_, do you mean calling `event.preventDefault()`? or are there other ways?

PR #165, opened 2021-09-08
Ignore the promise content on `transitionWhile()`

posva:
I think this change would make sense since the promise content is not used and without it things like this fail in TS:

```ts
function doAsyncStuff() {
  return fetchSomething().then(data => {
    storeData(data)
    return data // so it can be used with const data = await doAsyncStuff()
  })
}

event.transitionWhile(doAsyncStuff())
```

### DIFF
--- a/app_history.d.ts
+++ b/app_history.d.ts
@@ -100,7 +100,7 @@ declare class AppHistoryNavigateEvent extends Event {
   readonly formData: FormData|null;
   readonly info: unknown;
 
-  transitionWhile(newNavigationAction: Promise<void>): void;
+  transitionWhile(newNavigationAction: Promise<any>): void;
 }
 
 interface AppHistoryNavigateEventInit extends EventInit {


domenic:
Interesting. The only precedent I can find for this is https://github.com/microsoft/TypeScript/blob/07fd7bce642b5e3227137672fbaebb7762cfbc63/lib/lib.webworker.d.ts#L1542 which actually uses `any`, not even `Promise<any>`. I guess because (per Web IDL) you can pass anything in and it'll get converted to a promise...

https://github.com/microsoft/TypeScript/blob/07fd7bce642b5e3227137672fbaebb7762cfbc63/lib/lib.webworker.d.ts#L1570 is also somewhat relevant but there the type matters, so it can't help as much.

I'm happy to accept this, but let's leave it up for a day to see if any other folks who know TypeScript have opinions, since I am not an expert.

frehner:
>I guess because (per Web IDL) you can pass anything in and it'll get converted to a promise...

Interesting; will `transitionWhile` do the same thing? Or will it just immediately fulfill? Or will it throw if it's not an error?

domenic:
> Interesting; will transitionWhile do the same thing? Or will it just immediately fulfill? Or will it throw if it's not a promise?

Calling `transitionWhile(X)` will be equivalent to calling `transitionWhile(Promise.resolve(X))`, for any `X`.

PR #169, opened 2021-09-14
Disable most of the API on opaque origins

### LINKED ISSUE #167: Do opaque origins have access to the API
Related to #166, I wonder what kind of view we want to offer to opaque origins. Especially given that it seems we mainly want to disable storage APIs in them.

You could conceivably have multiple entries by having an opaque origin document create a blob URL and then navigate to that. Those would be same origin in principle. (This would be the first time that would be observable though, which kinda makes me think we shouldn't support it unless there's a use case.)

domenic:
Closes #167.

/cc @annevk @jakearchibald

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/169.html" title="Last updated on Sep 23, 2021, 7:15 PM UTC (cae4289)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/169/1d50439...cae4289.html" title="Last updated on Sep 23, 2021, 7:15 PM UTC (cae4289)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -190,6 +190,20 @@ dictionary AppHistoryResult {
 };
 </xmp>
 
+<div algorithm>
+  An {{AppHistory}} |appHistory| <dfn for="AppHistory">has entries and events disabled</dfn> if the following steps return true:
+
+  1. Let |browsingContext| be |appHistory|'s [=relevant global object=]'s [=Window/browsing context=].
+
+  1. If |browsingContext| is null, then return true.
+
+  1. If |browsingContext| is <a spec="HTML">still on its initial `about:blank` `Document`</a>, then return true.
+
+  1. If |appHistory|'s [=relevant global object=]'s [=associated Document=]'s [=Document/origin=] is [=opaque origin|opaque=], then return true.
+
+  1. Return false.
+</div>
+
 <h3 id="entries-api">Introspecting the app history entry list</h3>
 
 <dl class="domintro non-normative">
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -185,6 +185,20 @@ dictionary AppHistoryReloadOptions : AppHistoryNavigationOptions {
 };
 </xmp>
 
+<div algorithm>
+  An {{AppHistory}} |appHistory| <dfn for="AppHistory">has entries and events disabled</dfn> if the following steps return true:
+
+  1. Let |browsingContext| be |appHistory|'s [=relevant global object=]'s [=Window/browsing context=].
+
+  1. If |browsingContext| is null, then return true.
```
jakearchibald, line 193: When is the browsing context null?

jakearchibald: Ah, is it when the document isn't active?

jakearchibald (approved):

spec.bs
domenic: Yeah, exactly.

spec.bs
```diff
@@ -185,6 +185,20 @@ dictionary AppHistoryReloadOptions : AppHistoryNavigationOptions {
...
+
+  1. If |browsingContext| is <a spec="HTML">still on its initial `about:blank` `Document`</a>, then return true.
```
annevk, line 195: This is a bit surprising. It essentially inherits a bunch of state from the parent in this case. Is there something missing that would make supporting history problematic?

domenic: I think it's doable, but so far I've been saying that the initial about:blank is a scary place (e.g., most navigations from it are auto-converted into replacements; you can't navigate back to it currently; etc.) so it'd be best to avoid exposing that complexity.

annevk: It's easier to allow later than to disallow later so I'm okay with it, FWIW.

domenic:
https://bugs.chromium.org/p/chromium/issues/detail?id=1252490

PR #170, opened 2021-09-22
Don't fire navigate for non-fetch-scheme cases

domenic:
Notably, for javascript: URL navigations.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/170.html" title="Last updated on Sep 23, 2021, 7:08 PM UTC (7753d45)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/170/e01412f...7753d45.html" title="Last updated on Sep 23, 2021, 7:08 PM UTC (7753d45)">Diff</a>


domenic:
This also prohibits firing it for `about:srcdoc` navigations since those call the navigate algorithm with a response, not a request.

domenic:
I'm gonna tweak this to not ignore about:srcdoc after all, since those are a bit less weird. We can discuss blocking those separately.

domenic:
https://bugs.chromium.org/p/chromium/issues/detail?id=1252487 tracking issue

PR #189, opened 2021-11-17
Make "no-referrer" censor cross-document AppHistoryEntry URLs

### LINKED ISSUE #71: Scope app history entries to the browsing context group
Currently they're scoped to same-origin contiguous and same-frame. However we also need to account for browsing context group swaps due to COOP.

In the explainer this will be a minor sentence update or so, but we'll need caution when speccing.

domenic:
Also be clear that, despite the shaky spec foundations, the intent after this fix is to expose session history entries across browsing context group swaps.

Closes #71.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/189.html" title="Last updated on Nov 17, 2021, 9:22 PM UTC (f9d859d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/189/e8060a6...f9d859d.html" title="Last updated on Nov 17, 2021, 9:22 PM UTC (f9d859d)">Diff</a>

### DIFF
--- a/app_history.d.ts
+++ b/app_history.d.ts
@@ -59,7 +59,7 @@ interface AppHistoryEntryEventMap {
 declare class AppHistoryEntry extends EventTarget {
   readonly key: string;
   readonly id: string;
-  readonly url: string;
+  readonly url: string|null;
   readonly index: number;
   readonly sameDocument: boolean;
 
--- a/spec.bs
+++ b/spec.bs
@@ -35,6 +35,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: current entry; url: history.html#current-entry
     for: session history entry
       text: document; url: history.html#she-document
+      text: policy container; url: history.html#she-policy-container
       text: URL; url: history.html#she-url
     for: history handling behavior
       text: default; url: browsing-the-web.html#hh-default
@@ -49,6 +50,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
       text: discard; url: window-object.html#a-browsing-context-is-discarded
   type: method
     for: Document; text: open(unused1, unused2); url: multipage/dynamic-markup-insertion.html#dom-document-open
+  type: http-header
+    text: Cross-Origin-Opener-Policy; url: multipage/iana.html#cross-origin-opener-policy-2
 spec: html; urlPrefix: https://whatpr.org/html/6315/
   type: dfn
     text: traversable navigable; for: navigable; url: history.html#nav-traversable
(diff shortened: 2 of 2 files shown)


app_history.d.ts
```diff
@@ -59,7 +59,7 @@ interface AppHistoryEntryEventMap {
 declare class AppHistoryEntry extends EventTarget {
   readonly key: string;
   readonly id: string;
-  readonly url: string;
+  readonly url: string|null;
```
natechapin, line 62: The obvious thing in the code leaves the string empty rather than null. Catching that and converting it to null is presumably a 1-liner, but not the default.

natechapin (approved):

spec.bs
```diff
@@ -1419,7 +1430,9 @@ Each {{AppHistoryEntry}} has an associated <dfn for="AppHistoryEntry">index</dfn
 <div algorithm>
   The <dfn attribute for="AppHistoryEntry">url</dfn> getter steps are:
 
-  1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=Document/fully active=], then return the empty string.
+  1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=Document/fully active=], then return null.
+  1. Let |she| be [=this=]'s [=AppHistoryEntry/session history entry=].
+  1. If |she|'s [=session history entry/document=] does not equal [=this=]'s [=relevant global object=]'s [=associated Document=], and |she|'s [=session history entry/policy container=]'s [=policy container/referrer policy=] is <a>"`no-referrer`"</a>, then return null.
```
annevk, line 1435: Nit: no need for comma before and (while it looks like a lot of words, that's mostly in the source).

annevk (commented):
Okay, so the idea is that if you're on the document with no-referrer you get to see everything, but elsewhere you do not get to see the URL. But elsewhere you do get to see key/id/index/state? Is that what we want or should we similarly protect those?

domenic:
I don't think we want to protect those:

- key/id are browser-generated UUIDs and not sensitive information. Prohibiting access to key would also prohibit traversal back to that entry which seems suboptimal since it'd force people to using `history.go()`.
- index is trivially computable since `x.index` is just a convenience getter for to `appHistory.entries().indexOf(x)`
- state is something you affirmatively opt in to storing; see previous discussion in https://github.com/WICG/app-history/issues/71#issuecomment-854951702

annevk:
I was thinking you might want to have private state just for history entries on that document. I think I'm okay with not going to those lengths though.

Issue #190, opened 2021-11-18
Focus management for SPA navs


domenic:
This is spinning off from #25. It was also briefly mentioned in #162.

We have consistent feedback that focus management during SPA navs is an accessibility concern. [This writeup](https://www.gatsbyjs.com/blog/2019-07-11-user-testing-accessible-client-routing/) goes into good detail.

I think we should make sure that by default, same-document navigations that are controlled by app history (i.e., using `navigateEvent.transitionWhile()`) give a good focus experience. The starting point here is to get parity with cross-document navs (MPA navs). This means:

* For "push", "replace", and "reload" navigations, resetting focus to the `<body>` element, or the first `autofocus=""` element if there is one.
* For "traverse" navigations, it's complicated.

https://boom-bath.glitch.me/focus-navigation-test.html is a useful test page for this.

For traverse navigations, if the page is stored in the bfcache, then the current plan per https://github.com/whatwg/html/issues/5878 / https://github.com/whatwg/html/pull/6696 is to restore focus. I.e., since the entire `Document` is kept around, we just make sure not to clear its focus state when transitioning into or out of bfcache.

But for traverse navigations where the result is not coming from bfcache, there is no attempt at focus restoration in current browsers. I.e. there is nothing like scroll restoration (cf. #187) which tries to somehow find the right element to focus in this newly-created from-network-or-cache `Document`. It just does the usual `<body>`-or-`autofocus=""` dance (and `autofocus=""` doesn't seem to work in Chrome...).

So what should we do for app history and traverse navigations?

* On the one hand, SPA navs are supposed to be snappy and kinda like bfcache navs. So trying to make sure the focus is on the same element when you go back would be ideal.
* On the other hand, in the general case we won't be able to identify "the same element"!
  * In most SPA architectures, off-screen content (i.e. from a previous history entry) will have been at least removed from the DOM, and probably destroyed via garbage collection; when you traverse back, the page will reconstruct new DOM elements for the content. We could keep a pointer to the element around, but that'd be a potential memory leak, and most likely that specific element won't reenter the DOM anyway, so we wouldn't get anything out of it in most cases. The only case where this would work really well would be when an SPA is implemented by just hiding and showing things, which might be the case for some toy apps but doesn't seem likely in general.
[...]

Issue #187, opened 2021-11-16
Better control over scroll restoration during traversal


domenic:
Talking with a few web developers, we've pinpointed a few issues with scroll restoration during traversal, that seem relevant to any history/navigation API.

The essential problem we've heard so far is that scroll restoration happens unpredictably and often at the wrong times. For example:

- The browser tries to restore scroll position, but your app is still setting up the DOM and the relevant elements aren't ready yet
- The browser tries to restore scroll position, but the page's contents have changed and scroll restoration doesn't work that well (e.g., in a listing of files in some shared folder, someone deleted a bunch of the files)
- You need to make some measurements in order to do a proper transition, but the browser does scroll restoration during your transition which messes it up. ([Demo](https://nifty-blossom-meadow.glitch.me/legacy-history/transition.html))

Note that these problems are present for both cross-document traversals and same-document traversals.

The developers I've talked to have expressed that they're looking for something between the two extremes we currently have available:

- Disabling scroll restoration and handling it yourself (via `history.scrollRestoration = "manual"`)
- Letting the browser perform scroll restoration at an unpredictable time and in an unpredictable manner, after traversal

Here's an initial stab at a solution for this problem space. Note that the core proposal is not really tied to the rest of the app history API, and might end up advancing as a separate proposal on the [whatwg/html](https://github.com/whatwg/html) repo. But I wanted to start with this community, since it fits into the general problem space of "ideal new APIs for handling history/navigation". And I do have an app history specific bonus proposal below the main one.

#### Proposal

Introduce a new event, `beforescrollrestoration`. It can be targeted at any scrollable container, with the `<html>` element being the one for the overall viewport. (Or, the `<body>` element, in quirks mode...) But e.g. if you were implementing your page as one large scrollable `<div>` inside a non-scrolling viewport, then you'd listen for the event on that `<div>`.

This event would fire whenever the browser would normally perform scroll restoration. Per the [current spec](https://html.spec.whatwg.org/#restore-persisted-state) and [tests](https://github.com/web-platform-tests/wpt/blob/master/html/browsers/browsing-the-web/history-traversal/scroll-restoration-order.html), is after `popstate`/before `hashchange`. (App history's `currentchange` comes before `popstate`, so the total sequence woul be: `currentchange`, `popstate`, `beforescrollrestoration`, `hashchange`.) But, see below...

The event has the following methods:

- `preventDefault()`: stops scroll restoration from happening, for now
[...]

PR #197, opened 2022-01-07
Explainer for focus management and scroll restoration

domenic:
This also rearranges some of the content around accessibility technology announcements and loading spinners/stop buttons, grouping them all under a new section "Customizations and consequences of navigation interception".

Still a few TODOs for next week, thus leaving this in a draft state. But the plan is for this PR to address #187, #190, and #25.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/pull/197.html" title="Last updated on Jan 27, 2022, 10:43 PM UTC (8affd6a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/app-history/197/ff79291...8affd6a.html" title="Last updated on Jan 27, 2022, 10:43 PM UTC (8affd6a)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,9 +72,13 @@ for (const entry of performance.getEntriesByType("same-document-navigation")) {
     - [Example: async transitions with special back/forward handling](#example-async-transitions-with-special-backforward-handling)
     - [Example: progressively enhancing form submissions](#example-progressively-enhancing-form-submissions)
     - [Restrictions on firing, canceling, and responding](#restrictions-on-firing-canceling-and-responding)
-    - [Accessibility benefits of standardized single-page navigations](#accessibility-benefits-of-standardized-single-page-navigations)
     - [Measuring standardized single-page navigations](#measuring-standardized-single-page-navigations)
     - [Aborted navigations](#aborted-navigations)
+  - [Customizations and consequences of navigation interception](#customizations-and-consequences-of-navigation-interception)
+    - [Accessibility technology announcements](#accessibility-technology-announcements)
+    - [Loading spinners and stop buttons](#loading-spinners-and-stop-buttons)
+    - [Focus management](#focus-management)
+    - [Scroll position restoration](#scroll-position-restoration)
   - [Transitional time after navigation interception](#transitional-time-after-navigation-interception)
     - [Example: handling failed navigations](#example-handling-failed-navigations)
     - [Example: single-page app redirects and guards](#example-single-page-app-redirects-and-guards)
@@ -494,16 +498,6 @@ Finally, the following navigations **cannot be replaced with same-document navig
 
 We'll note that these restrictions still allow canceling cross-origin non-back/forward navigations. Although this might be surprising, in general it doesn't grant additional power. That is, web developers can already intercept `<a>` `click` events, or modify their code that would set `location.href`, even if the destination URL is cross-origin.
 
-#### Accessibility benefits of standardized single-page navigations
-
-The `navigate` event's `event.transitionWhile()` method provides a helpful convenience for implementing single-page navigations, as discussed above. But beyond that, providing a direct signal to the browser as to the duration and outcome of a single-page navigation has benefits for accessibility technology users.
-
-In particular, with [cross-document](#appendix-types-of-navigations) navigations, AT users get clear feedback that a navigation has occurred. But traditionally, single-page navigations have not been communicated in the same way to accessibility technology. This is in part because it is not clear to the browser when a user interaction causes a single-page navigation, because of the app-specific JavaScript that intermediates between such interactions and the eventual call to `history.pushState()`/`history.replaceState()`. In particular, it's unclear exactly when the navigation begins and ends: trying to use the URL change as a signal doesn't work, since when applications call `history.pushState()` during the content loading process varies.
-
-Implementing single-page navigations by using the `navigate` event and its `transitionWhile()` function solves this part of the problem. It gives the browser clear insight into when a navigation is being handled as a single-page navigation, and the provided promise allows the browser to know how long the navigation takes, and whether or not it succeeds. We expect browsers to use these to update their own UI (including any loading indicators; see [whatwg/fetch#19](https://github.com/whatwg/fetch/issues/19) and [whatwg/html#330](https://github.com/whatwg/html/issues/330) for previous feature requests). And we expect browsers to communicate these signals to accessibility technology, in the same way they do for traditional cross-document navigations.
-
-This does not yet solve all accessibility problems with single-page navigations. In particular, this proposal doesn't currently have a solution for focus management and placing the user's keyboard focus in the relevant place after navigation. However, we are very interested in finding a way to make usage of the app history API guide web developers toward creating accessible experiences, and would like to explore additions or changes that would help with these aspects of the problem as well. Please join us to discuss and brainstorm in [#25](https://github.com/WICG/app-history/issues/25).
-
 #### Measuring standardized single-page navigations
 
 Continuing with the theme of `transitionWhile()` giving ecosystem benefits beyond just web developer convenience, telling the browser about the start time, duration, end time, and success/failure if a single-page app navigation has benefits for metrics gathering.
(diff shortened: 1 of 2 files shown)


domenic:
Hmm, `transitionWhile()` being the place to set these options runs into unclear semantics if you set it twice:

```js
  e.transitionWhile(Promise.resolve(), { focusReset: "manual" });
  e.transitionWhile(Promise.resolve(), { focusReset: "after-transition" });
});
```

Options I see...

- Require the options to always match, throwing an exception on callers after the first if they mismatch. (Not great for apps trying to use multiple `transitionWhile()` calls or `navigate` event handlers in a non-coordinated fashion.)

- Try to come up with a conflict-resolution procedure. E.g.:
  - First wins (with later values ignored)
  - Last wins (with previous values ignored)
  - if `"manual"` is ever specified (i.e., not the default), it wins over `"after-transition"`. (Gets complicated if we ever expand the value space like the explainer indicates we might.)
  - Any of these could include console warnings, I guess.

- Abandon the idea of tying these options to `transitionWhile()` calls? E.g. `navigateEvent.setOptions(...)`?
  - It's not clear this really solves the problem, since you could still call that method twice. But it might give us flexibility to pick one of the above strategies in a way that's less confusing.
  - This is a bit less appealing because these options generally only make sense in the `transitionWhile()` case anyway... we'd basically duplicate all of the precondition logic.
  - Maybe something global, like `appHistory.setTransitionOptions(...)` or `appHistory.focusReset = ...`, `appHistory.scrollRestoration = ...`?

domenic:
I think the primary question is: do we expect people to want the same values for `focusReset` / `scrollRestoration` for all app history-mediated navigations in their app? Or might it vary depending on the navigation in question?

If we expect it to be the same, then a global API makes sense. Otherwise, it'd need to be configured per-`navigate` event.

README.md
```diff
@@ -547,6 +541,92 @@ In this case:
...
+
+We can also extend the `focusReset` option with other behaviors in the future. Here are a couple which have been proposed, but we are not planning to include in the initial version unless we get strong developer feedback that they would be helpful:
+
+- `e.transitionWhile(promise, { focusReset: "immediate" })`: immediately resets the focus to the `<body>` element, without waiting for the promise to settle.
+- `e.transitionWhile(promise, { focusReset: "two-stage" })`: immediately resets the focus to the `<body>` element, and then has the same behavior as `"after-transition"`.
+
+#### Scroll position restoration
+
+A common paint point for web developers is scroll restoration during traversal (back/forward) navigations. The essential problem is that scroll restoration happens unpredictably, and often at the wrong times. For example:
```
tbondwilkinson, line 592: "A common paint point"

Should be pain

README.md
```diff
@@ -547,6 +541,92 @@ In this case:
...
+
+- For traversals (i.e. cases where `navigateEvent.navigationType === "traverse"`), getting parity with the [back/forward cache experience](https://web.dev/bfcache/) requires restoring focus to the same element that was previously focused when that history entry was active. Unfortunately, this isn't something the browser can do automatically for client-side rendered applications; the notion of "the same element" [is not generally stable](https://github.com/WICG/app-history/issues/190#:~:text=On%20the%20other%20hand%2C%20in%20the%20general%20case%20we%20won%27t%20be%20able%20to%20identify%20%22the%20same%20element%22!) in such cases. So for such cases, using `focusReset: "manual"`, storing some identifier for the currently-focused element in the app history state, and calling `element.focus()` appropriately upon transition could give a better experience.
```
tbondwilkinson, line 581: Can you comment here when it would be appropriate to call element.focus() after transition? Would you recommend using the transition property, or in a currentchange listener, after the promise resolves, etc.

README.md
```diff
@@ -547,6 +541,92 @@ In this case:
...
+- The application needs to perform some measurements in order to do a proper transition, but the browser does scroll restoration during the transition, which messes up those measurements. ([Demo of this problem](https://nifty-blossom-meadow.glitch.me/legacy-history/transition.html): notice how when going back to the grid view, the transition sends the square to the wrong location.)
+
+Currently the browser provides two options: performing scroll restoration automatically, or disabling it entirely with `history.scrollRestoration = "manual"`. App history gives us an opportunity to provide some intermediate options to developers, at least for the case of same-document transitions. We do this via another option to `transitionWhile()`:
+
+- `e.transitionWhile(promise, { scrollRestoration: "after-transition" })`: the default behavior. The browser delays its scroll restoration logic until `promise` fulfills; it will perform no scroll restoration if the promise rejects. If the user has scrolled during the transition then no scroll restoration will be performed.
```
tbondwilkinson, line 600: Do we really want to cancel scroll behavior if the user scrolls during the transition? I feel like the current behavior in a multi-page application is to disregard user scrolls while the page is navigating?

README.md
```diff
@@ -547,6 +541,92 @@ In this case:
...
+- `e.transitionWhile(promise, { scrollRestoration: "manual" })`: The browser will perform no automatic scroll restoration. However, the developer can use the below API to get semi-automatic scroll restoration.
+
+When using `scrollRestoration: "manual"`, the `e.restoreScroll()` API is available. This will perform the browser's scroll restoration logic at the specified time. This allows cases that require precise control over scroll restoration timing, such as a non-broken version of the [demo referenced above](https://nifty-blossom-meadow.glitch.me/legacy-history/transition.html), to be written like so:
+
+```js
+navigateEvent.transitionWhile((async () => {
+  await fetchDataAndSetUpDOM();
+  navigateEvent.restoreScroll();
```
tbondwilkinson, line 608: Do we want the navigateEvent to be long-lasting? I remember there being issues maybe a long time ago with Events that had a lifetime, where the browser tried to get rid of them but they were still around. I suppose the only other option would be to hang this API off appHistory itself.

README.md
```diff
@@ -547,6 +541,92 @@ In this case:
...
+
+- `restoreScroll()` will silently do nothing if called after the user has started scrolling the document.
+
+- `restoreScroll()` doesn't actually perform a single update of the scroll position. Rather, it puts the page in scroll-position-restoring mode. The scroll position could update several times as more elements load and [scroll anchoring](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor/Guide_to_scroll_anchoring) kicks in.
+
+- By default, any navigations which are intercepted with `e.transitionWhile()` will _ignore_ the value of `history.scrollRestoration` from the classic history API. This allows developers to use `history.scrollRestoration` for controlling cross-document scroll restoration, while using the more-granular option to `transitionWhile()` to control individual same-document navigations.
+
+We could also add the following APIs in the future, but we are currently not planning on including them until we hear developer feedback that they'd be helpful:
+
+- `e.transitionWhile(promise, { scrollRestoration: "immediate" })`: the browser performs its usual scroll restoration logic, but does so immediately instead of waiting for `promise`.
```
tbondwilkinson, line 625: This is equivalent to calling restoreScroll immediately though, right?

README.md
```diff
@@ -547,6 +541,92 @@ In this case:
...
+- `e.transitionWhile(promise, { scrollRestoration: "auto" })`: the browser performs its usual scroll restoration logic, at its usual indeterminate time.
```
tbondwilkinson, line 626: Would anyone want this? Sounds dubious.

README.md
```diff
@@ -547,6 +541,92 @@ In this case:
...
+
+An additional API that would be helpful, both for cases like these and more generally, would be one for [setting and getting the sequential focus navigation start point](https://github.com/whatwg/html/issues/5326). Especially for the custom traversals case, this would give even higher-fidelity focus restoration. (But that proposal is separate from app history.)
+
+We can also extend the `focusReset` option with other behaviors in the future. Here are a couple which have been proposed, but we are not planning to include in the initial version unless we get strong developer feedback that they would be helpful:
+
+- `e.transitionWhile(promise, { focusReset: "immediate" })`: immediately resets the focus to the `<body>` element, without waiting for the promise to settle.
+- `e.transitionWhile(promise, { focusReset: "two-stage" })`: immediately resets the focus to the `<body>` element, and then has the same behavior as `"after-transition"`.
+
+#### Scroll position restoration
```
tbondwilkinson, line 590: I'm wondering about how there's no ability to influence WHERE the browser scrolls. This just provides an API for WHEN the browser restores scroll position. is that correct?

README.md
dvoytenko, line 581: IMHO heuristics with (1) the input object instance (if it wasn't removed from DOM), and (2) ID should be a good start. Are there any other similar restoration rules, e.g. from the current BFCache heuristics? Maybe also include an input's name? ID are often autogenerated, but names are usually logical.

README.md
dvoytenko: I'd tend to agree: user scroll should be ignored. But this rule could apply to manual calls to `window.scrollTo` and similar APIs.

README.md
```diff
@@ -547,6 +541,92 @@ In this case:
...
+
+- `e.transitionWhile(promise, { scrollRestoration: "after-transition" })`: the default behavior. The browser delays its scroll restoration logic until `promise` fulfills; it will perform no scroll restoration if the promise rejects. If the user has scrolled during the transition then no scroll restoration will be performed.
+- `e.transitionWhile(promise, { scrollRestoration: "manual" })`: The browser will perform no automatic scroll restoration. However, the developer can use the below API to get semi-automatic scroll restoration.
+
+When using `scrollRestoration: "manual"`, the `e.restoreScroll()` API is available. This will perform the browser's scroll restoration logic at the specified time. This allows cases that require precise control over scroll restoration timing, such as a non-broken version of the [demo referenced above](https://nifty-blossom-meadow.glitch.me/legacy-history/transition.html), to be written like so:
```
dvoytenko, line 603: I'm wondering if we're going down the manual route, should we just leave it to the app to restore scroll manually? `e.restoreScroll()` is otherwise useful because an app doesn't have to save an old scroll position manually. But would this make API excessively complicated?

README.md
```diff
@@ -547,6 +541,92 @@ In this case:
...
+
+- By default, any navigations which are intercepted with `e.transitionWhile()` will _ignore_ the value of `history.scrollRestoration` from the classic history API. This allows developers to use `history.scrollRestoration` for controlling cross-document scroll restoration, while using the more-granular option to `transitionWhile()` to control individual same-document navigations.
+
+We could also add the following APIs in the future, but we are currently not planning on including them until we hear developer feedback that they'd be helpful:
+
+- `e.transitionWhile(promise, { scrollRestoration: "immediate" })`: the browser performs its usual scroll restoration logic, but does so immediately instead of waiting for `promise`.
+- `e.transitionWhile(promise, { scrollRestoration: "auto" })`: the browser performs its usual scroll restoration logic, at its usual indeterminate time.
+- `const { x, y } = e.scrollDestination()` giving the current position the browser would restore to, if `e.restoreScroll()` was called.
```
dvoytenko, line 627: If we have this API, I think we can drop `e.restoreScroll()` for manual mode.

README.md
domenic: Any of `appHistory.transition.finished.then(updateFocus)`, `appHistory.back().then(updateFocus)` or `appHistory.addEventListener("navigatesuccess", updateFocus)` would work; they have the same timing.

If you wanted to update focus even on failure then you'd do `appHistory.transition.finished.finally(updateFocus)` or `appHistory.back().finally(updateFocus)`. For the event version you'd add handlers for both `navigatesuccess` and `navigateerror`.

You could also integrate it into your specific application logic, e.g.

```js
  e.transitionWhile((async () => {
    await updateDOMWithSkeleton();
    findElement().focus();
    await updateTheRestOfTheDOM();
  })());
});
```

I'll mention the basics here in the PR.

README.md
domenic: The Chromium engineers pointed me to code which abandons scroll restoration (for MPA navs) when the user scrolls. I will do some black-box testing to confirm.

domenic: Here is an example:

1. Open https://neat-equal-cent.glitch.me/, preferably in a vertically-small-ish window
2. Wait for, e.g., paragraph 50 to load, and scroll down to it
3. Click the example.com link
4. Click back. Do not touch anything.
5. After paragraph 50 loads, scroll position will be restored.

Then try it this way:

4. Click back. Once a scrollbar appears, scroll slightly.
5. Even after paragraph 50 loads, your scroll position will not be disturbed. You having scrolled will have changed how it works.

README.md
domenic: It should be OK. At least for the duration of the navigation, I believe we keep it around anyway.

README.md
domenic: Yep.

README.md
domenic: Right. You can use `window.scrollTo()` + `scrollRestoration: "manual"` for manual scrolling. I can mention that...

README.md
domenic: Well, that's why it's in the unclear-utility section :).

README.md
domenic: I'm hopeful `e.restoreScroll()` is doable without much work, so I kept it in the explainer as an MVP feature. It's true we could omit it though without any compat issues, so it might get deprioritized if other things come up or if we find out that it's not as easy to implement as I'd hoped.

domenic:
Offline feedback is that it might vary depending on navigation. So we still have lots of options to choose from in how to resolve conflicts...

I think I'd prefer to tie this to `transitionWhile()`. So the alternatives are:

- Throw on mismatches
- Last wins
- First wins
- Something more complicated

I think throwing is too painful for people who want to have multiple uncoordinated transitionWhile() handlers; see #94 for the background there. Although it should be rare and might not work perfectly, it shouldn't be almost-always-broken.

So I'm going to propose two alternatives:

1. Last wins

2. More-complicated version

   In the more-complicated version, `transitionWhile(..., { focusReset: "after-transition" })` is *not* equivalent to `transitionWhile(..., { focusReset: undefined })` or `transitionWhile(...)`. So we have three possibilities: `"after-transition"`, `"manual"`, and `undefined`. The rule is then the last non-`undefined` value wins.
   
   The intent here is that if we have multiple people calling `transitionWhile()`, ones that don't explicitly set `focusReset` don't impact the outcome, even if they were called after cases that did set it.

I'll spec last-wins for now but am very open to feedback as to whether the more-complicated version, or first wins, would be better.

domenic:
Going to split the spec work out into its own PR so I can keep noodling, and then we can point people to the actual README for the current plans.

PR #218, opened 2022-03-23
Specify navigateEvent.downloadRequest

### LINKED ISSUE #76: <a download> is a navigation; how does it interact with the navigation API?
Background: https://github.com/whatwg/html/issues/5548, which is basically about how `<a download>` is a type of navigation, even though the specs currently don't treat it as such.

This means the navigation API should have a well-defined interaction with such downloads, in the spec and tests. Speccing this properly might require fixing https://github.com/whatwg/html/issues/5548.

My instinct is that such downloads should not trigger the navigation API's `navigate` event.

domenic:
Closes #76.

@natechapin, please review!

@smaug---- @annevk any review you have time to provide would help validate that this is precise enough to be implemented interoperably, despite building on shaky foundations.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/navigation-api/pull/218.html" title="Last updated on Mar 25, 2022, 4:05 PM UTC (aaff835)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/navigation-api/218/4a9fdaa...aaff835.html" title="Last updated on Mar 25, 2022, 4:05 PM UTC (aaff835)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ An application or framework's centralized router can use the `navigate` event to
 
 ```js
 navigation.addEventListener("navigate", e => {
-  if (!e.canTransition || e.hashChange) {
+  if (!e.canTransition || e.hashChange || e.requestDownload !== null) {
     return;
   }
 
@@ -323,6 +323,8 @@ The event object has several useful properties:
 
 - `formData`: a [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) object containing form submission data, or `null` if the navigation is not a form submission.
 
+- `requestDownload`: a string or null, indicating whether this navigation was initiated by a `<a href="..." download>` link. If it was, then this will contain the value of the attribute (which could be the empty string).
+
 - `info`: any value passed by `navigation.navigate(url, { state, info })`, `navigation.back({ info })`, or similar, if the navigation was initiated by one of those methods and the `info` option was supplied. Otherwise, undefined. See [the example below](#example-using-info) for more.
 
 - `signal`: an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) which can be monitored for when the navigation gets aborted.
@@ -358,8 +360,8 @@ navigation.addEventListener("navigate", e => {
     return;
   }
 
-  // Don't intercept fragment navigations.
-  if (e.hashChange) {
+  // Don't intercept fragment navigations or downloads.
+  if (e.hashChange || e.requestDownload !== null) {
     return;
   }
 
@@ -405,7 +407,7 @@ Sometimes it's desirable to handle back/forward navigations specially, e.g. reus
 ```js
 navigation.addEventListener("navigate", e => {
   // As before.
-  if (!e.canTransition || e.hashChange) {
+  if (!e.canTransition || e.hashChange || e.requestDownload !== null) {
     return;
   }
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ An application or framework's centralized router can use the `navigate` event to
 
 ```js
 navigation.addEventListener("navigate", e => {
-  if (!e.canTransition || e.hashChange || e.requestDownload !== null) {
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -405,7 +407,7 @@ Sometimes it's desirable to handle back/forward navigations specially, e.g. reus
 ```js
 navigation.addEventListener("navigate", e => {
   // As before.
-  if (!e.canTransition || e.hashChange) {
+  if (!e.canTransition || e.hashChange || e.requestDownload !== null) {
```
domenic, line 410: I wish there were a way to make this truthy for `<a download href="...">`. I feel like `if (!e.requestDownload)` might be a common mistake, which such no-filename `download=""`s would trigger.

The only things I can think of are:
- Use a sentinel filename like `"unknown"` for the `<a download>` case. (Ick.)
- Use two fields, `downloadRequested` and `downloadRequestFilename`. (Feels like a lot of API surface for such an edge case.)

spec.bs
```diff
@@ -1092,6 +1102,20 @@ enum NavigationNavigationType {
...
+    <p>Represents whether or not this navigation was requested to be a download, by using an <{a}> or <{area}> element's <{a/download}> attribute:
+
+    * If a download was not requested, then this property is null.
+    * If a download was requested, returns the filename that was supplied, via `<a download="filename" href="...">`. (This could be the empty string, as in the case of `<a download href="...">`.)
+
+    <p>Note that a download being requested does not always mean that a download will happen: for example, the download might be blocked by browser security policies, or end up being treated as a push navigation for <a href="https://github.com/whatwg/html/issues/7718" class="XXX">unspecified reasons</a>.
+
+    <p>Similarly, a navigation might end up being a download even if it was not requested to be one, due to the destination server responding with a `Content-Disposition: attachment` header.
+
+    <p>Finally, note that the {{Navigation/navigate}} event will not fire for at all for downloads initiated using browser UI affordances, e.g. those created by right-clicking and choosing to save the target of the link.
```
annevk, line 1116:
```suggestion
    <p>Finally, note that the {{Navigation/navigate}} event will not fire at all for downloads initiated using browser UI affordances, e.g., those created by right-clicking and choosing to save the target of the link.
```

annevk (commented):
This looks like a reasonable summary to me.

As for the API, maybe we could expose other getter(s) that group things folks typically want to filter these events on? Might be easier to judge once there's a bunch of code in the wild.

navigation_api.d.ts
```diff
@@ -118,6 +118,7 @@ declare class NavigateEvent extends Event {
   readonly destination: NavigationDestination;
   readonly signal: AbortSignal;
   readonly formData: FormData|null;
+  readonly requestDownlaod: string|null;
```
natechapin, line 121: downloadRequest

spec.bs
```diff
@@ -1216,6 +1240,7 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
   To <dfn>fire a traversal `navigate` event</dfn> at a {{Navigation}} |navigation| given a [=session history entry=] <dfn for="fire a traversal navigate event">|destinationEntry|</dfn>, and an optional [=user navigation involvement=] <dfn for="fire a traversal navigate event">|userInvolvement|</dfn> (default "<code>[=user navigation involvement/none=]</code>"):
 
   1. Let |event| be the result of [=creating an event=] given {{NavigateEvent}}, in |navigation|'s [=relevant Realm=].
+  1. Set |event|'s [=NavigateEvent/classic history API serialized data=] to null.
```
natechapin, line 1243: I don't immediately see why this is necessary?

natechapin (approved):

spec.bs
domenic: This is kind of an unrelated bug fix where I'm trying to initialize all the fields of NavigateEvent every time I create it. Previously the language around classic history API serialized data was imprecisely talking about being "used conditionally" and then assuming that meant it was OK to not initialize it.

PR #219, opened 2022-03-24
Change { replace: true } to { history: "replace" }

### LINKED ISSUE #111: Behavior of navigate(currentURL, { replace })
@jakearchibald, @annevk, @smaug---- and I discussed the behavior of `navigation.navigate(appHistory.current.url)` today.

The current spec says that this always does a replace navigation, similar to `location.href = location.href`. This is a bit weird because it means that the `replace` parameter is ignored, e.g.

```js
// All three of these are equivalent: they all do a replace.
navigation.navigate(navigation.currentEntry.url);
navigation.navigate(navigation.currentEntry.url, { replace: true });
navigation.navigate(navigation.currentEntry.url, { replace: false });
```

We discussed a few options for what to do:

1. Do a same-document navigation, either push or replace according to the `replace` option.
    - Precedent: `pushState(null, "")` / `replaceState(null, "")`
    - But, this is not great because in general we want `navigate()` to be about new-document navigations; to make them same-document, you intercept with the `navigate` event.
2. Do a new-document replace navigation, ignoring the `replace` option
    - Precedent: `location.href = location.href`, or `location.assign(location.href)`.
    - This is the spec's current behavior.
3. Do a new-document navigation, either push or replace according to the `replace` option.
    - A new-document push navigation to the same URL as the current one is only possible today by using redirects, and only in some browsers.
4. (3), but have the default be a somewhat-magic "replace for same-URL, push for different-URL". I.e.

    ```js
    navigation.navigate(navigation.currentEntry.url);                      // does a replace
    navigation.navigate(navigation.currentEntry.url, { replace: true });   // does a replace
    navigation.navigate(navigation.currentEntry.url, { replace: false });  // does a push

    navigation.navigate(someOtherURL);                                // does a push
    navigation.navigate(someOtherURL, { replace: true });             // does a replace
    navigation.navigate(someOtherURL, { replace: false });            // does a push
    ```

5. Return a rejected promise whenever you pass `replace: false` and try to navigate to the current URL.

6. Return a rejected promise whenever you pass anything besides `replace: true` and try to navigate to the current URL. I.e. you need to be super-explicit.

We were leaning toward (4), although the lack of good precedent is a bit tricky.

domenic:
We can then introduce { history: "push" } which explicitly errors if we're in a convert-to-replace situation. Closes #111.

This also fixes a preexisting omission where we were not properly doing replace navigations before the document is loaded.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/navigation-api/pull/219.html" title="Last updated on Mar 24, 2022, 3:48 PM UTC (3c59d09)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/navigation-api/219/4a9fdaa...3c59d09.html" title="Last updated on Mar 24, 2022, 3:48 PM UTC (3c59d09)">Diff</a>


spec.bs
```diff
@@ -787,7 +810,7 @@ An <dfn>navigation API method navigation</dfn> is a [=struct=] with the followin
 
   1. Let |info| be |options|["{{NavigationOptions/info}}"] if it exists; otherwise, undefined.
 
-  1. Let |historyHandling| be "<a for="history handling behavior">`replace`</a>" if |options|["{{NavigationNavigateOptions/replace}}"] is true; otherwise, "<a for="history handling behavior">`default`</a>".
+  1. Let |historyHandling| be "<a for="history handling behavior">`replace`</a>" if |options|["{{NavigationNavigateOptions/history}}"] is "{{NavigationHistoryBehavior/replace}}" or if |document| is not <a spec="HTML">completely loaded</a>; otherwise, "<a for="history handling behavior">`default`</a>".
```
domenic, line 813: This is somewhat of a bugfix, in the sense that Chromium definitely already did this conversion, and I believe we have tests for it.

In Chromium, by default all before-load navigations are replacements, with a special exception for anchor clicks: see https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/frame/local_frame.cc;l=454;drc=a24520a700e2581395920c74d59a918aa60bd5c9 . However in the spec, each call site of the navigate algorithm independently changes historyHandling, except anchor clicks which do not.

Issue #221, opened 2022-03-29
Tracking issue for rebasing on top of the navigation and session history rewrite


domenic:
https://github.com/whatwg/html/pull/6315 rewrites much of the foundational standard that this spec is built on top of.

The eventual goal (on the order of a month or two) is to merge that PR into HTML, and then merge this spec into HTML. In the meantime, we need to make sure it's relatively clear how to implement this spec.

The current status is that this spec is built on a mixture of the live HTML Standard, an older revision of the session history rewrite, and (after #220) the latest revision of the session history rewrite. This is a pretty confusing state of affairs.

We can mitigate that via two parallel routes:

- Mitigation
  - Have discussion and scratch-notes on this issue about how things are supposed to fit together, despite the spec not being clear
  - Write lots of web platform tests
- Do the work
  - I.e., continue rebasing this spec on top of the session history rewrite, until it is completely rebased
  - This process will probably compete for resources with the project of landing the rewrite itself, and then migrating this spec to HTML

I am relatively confident that despite the current spec situation being subpar, it will not lead to new interop problems. I.e., I don't think there is anything about the rebasing that will require significant cross-browser discussion. There are some things in the rewrite themselves, at the foundational level, that are not interoperable, and will benefit from discussion, but how this spec fits on top of the rewrite is, in my judgment, pretty clear from the current spec text.

For example, the exact structure of a "session history entry" will change between the live HTML Standard and the session history rewrite. But, the idea that it needs to be [extended with new fields](https://wicg.github.io/navigation-api/#session-history-new-she-fields) will work the same before and after.

Similarly, in the rewrite, reloading is changing from using the navigate algorithm, to using a separate "reload" algorithm. But, it is pretty clear that reloading will in either case synchronously fire a `navigate` event according to the current spec's algorithms.

If anyone finds areas of the current Navigation API spec unclear in how they are supposed to fit into existing/rewrite infrastructure, please highlight them on this issue and I will fix them ASAP.

domenic:
Scratch notes about how to do the rebase (will be updated over time as I discover things):

(Here "main thread"/"renderer process" = tasks queued on a `Window` global; "browser process" = the [session history traversal queue](https://whatpr.org/html/6315/history.html#tn-session-history-traversal-queue))

[...]

PR #222, opened 2022-03-29
Rebase more on top of session history rewrite

domenic:
Similar to the previous commit, this attempts to tackle a specific area, not the whole rebase project (which is tracked in #221). It focuses on traversal, in particular integration with the new versions of "apply the history step" and "traverse the history by a delta".

This removes a TODO in the former definition of "perform a navigation API traversal".

@jakearchibald, can you review this? I have a particular question which I will leave inline on the diff.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/navigation-api/pull/222.html" title="Last updated on Mar 29, 2022, 7:39 PM UTC (2633cf9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/navigation-api/222/f06f2dc...2633cf9.html" title="Last updated on Mar 29, 2022, 7:39 PM UTC (2633cf9)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -65,6 +65,16 @@ spec: html; urlPrefix: https://whatpr.org/html/6315/
     text: reactivate; url: history.html#reactivate-a-document
     text: navigate to a fragment; url: history.html#navigate-fragid
     text: navigation and traversal task source; url: webappapis.html#navigation-and-traversal-task-source
+    text: node navigable; url: browsers.html#node-navigable
+    text: append the following session history traversal steps; url: history.html#tn-append-session-history-traversal-steps
+    text: snapshot source snapshot params; url: history.html#snapshotting-target-snapshot-params
+    text: traverse the history by a delta; url: browsers.html#traverse-the-history-by-a-delta
+    text: top-level traversable; url: history.html#top-level-traversable
+    for: apply the history step
+      text: checkForUserCancellation; url: history.html#apply-history-step-check
+      text: unsafeNavigationStartTime; url: history.html#apply-history-step-start-time
+      text: sourceSnapshotParams; url: history.html#apply-history-step-source-snapshot
+      text: initiatorToCheck; url: history.html#apply-history-step-initiator
     for: URL and history update steps
       text: serializedData; url: history.html#uhus-serializeddata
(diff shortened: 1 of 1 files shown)


domenic:
Unfortunately Jake seems to have gone out of office before having a chance to review this. I will merge it since it's definitely an improvement, but I still would like it if he could double-check when he gets back.

spec.bs
```diff
@@ -1011,25 +1025,13 @@ An <dfn>navigation API method navigation</dfn> is a [=struct=] with the followin
 
        <p class="note">This can occur if a previously-queued-up traversal already took us to this session history entry. In that case that previous traversal will have dealt with |ongoingNavigation| already.
 
-    1. Let |targetStep| be null.
+    1. [=Apply the history step=] given by |targetEntry|'s [=session history entry/step=] to |traversable|, with <i>[=apply the history step/checkForUserCancellation=]</i> set to true, <i>[=apply the history step/unsafeNavigationStartTime=]</i> set to |unsafeNavigationStartTime|, <i>[=apply the history step/sourceSnapshotParams=]</i> set to |sourceSnapshotParams|, and <i>[=apply the history step/initiatorToCheck=]</i> set to |navigable|.
```
jakearchibald, line 1028: Before this, should there be a check that the target entry is within the same contiguous same-origin block as the source? The intent of the API is to limit navigations & visibility to those entries right? It seems possible that another navigation may have changed those conditions

jakearchibald (commented):
Just one question

spec.bs
domenic: Interesting. The only way that could happen would be via the replace-splitting case, I think. But yeah, it seems bad, and deterministically triggerable. Let me file a new issue.

domenic: Hmm, I tried writing it up and I can't actually figure out how to trigger it. To do the replace navigation you have to do an actual full cross-document navigation. But then when you go back to a page which intends to do `traverseTo(nowInvalidKey)`, that page will get the updated `entries()` list, so the early checks inside `traverseTo()` should catch it.

In other words I can't figure out how to split the history during the period between `traverseTo()` being called, and the traversal hitting the traversal queue.

Issue #40, opened 2021-02-17
Maybe allow respondWith to take a promise-returning function?


domenic:
A lot of code examples I'm writing are of the sort

```js
appHistory.addEventListener("navigate", e => {
  e.respondWith((async () => {
    // Do actual async stuff here
  })());
});
```

It'd be nicer if this could be written as

```js
appHistory.addEventListener("navigate", e => {
  e.respondWith(async () => {
    // Do actual async stuff here
  });
});
```

with the exact same semantics.

On the other hand, we haven't done this for other promise-accepting APIs on the platform (e.g. service worker's `waitUntil()` and `respondWith()`). Probably it's a bit too magic. (I'd be interested to hear if there were historical deliberations about this; /cc @jakearchibald @annevk.)

The real solution here is [async do expressions](https://docs.google.com/presentation/d/1oPLCP0z7Bu9WMxnpM1jevtijGLHxp8_PuLyD3y2E8MA/edit#slide=id.g82a8a74c53_0_58) in the language (/cc @bakkot):

```js
appHistory.addEventListener("navigate", e => {
  e.respondWith(do async {
    // Do actual async stuff here
  });
});
```

but `do` expressions seem to be one of those features which is really bad at getting through TC39's consensus model so I'm not sure it's a good idea to wait for them.

bakkot:
> but `do` expressions seem to be one of those features which is really bad at getting through TC39's consensus model

I am hopeful that they'll advance at the next meeting (March 9), time permitting, or if time does not permit at the following one (April 19, the dates are weird because we tweaked the schedule this year). I have a concrete change to make (allowing `return` and `break`) which I believe will address the one expressed blocking concern, and from an informal survey of delegates there was more support for that change than there was opposition to it (and no one who said they'd block).

If `do` fails again at that meeting I'll probably try to advance `async do` on its own, since it avoids some of the thornier parts of `do`: in particular it cannot possibly `return` the outer function.

tbondwilkinson:
I know it's "old" but there's always `new Promise()` 

```
appHistory.addEventListener("navigate", e => {
  e.respondWith(new Promise((resolve) => {
     
  }));
});
```

And truthfully I don't know how much people will want to be using inline things here. They'll probably call out to some method.

```
appHistory.addEventListener("navigate", e => {
  e.respondWith(router.resondWith(e));
});
```

domenic:
[...]

PR #235, opened 2022-06-02
Replace transitionWhile()/canTransition with intercept()/canIntercept

### LINKED ISSUE #230: transitionWhile doesn't quite work for scroll restoration
A regular navigation happens roughly like this:

1. Begin fetching
1. Time passes
1. Fetch complete
1. Capture & store scroll position
1. Switch to new page
1. Restore scrolling if applicable

`transitionWhile` doesn't really fit in this model.

If an unsettled promise is passed to `transitionWhile` (due to pending network activity), it goes like this:

1. Begin fetching
1. `transitionWhile` called - capture & store scroll position
1. Time passes
1. Fetch complete
1. Alter DOM
1. Promise passed to `transitionWhile` resolves
1. Restore scrolling if applicable

This isn't ideal, as the user may have scrolled during the fetch, but that change is discarded.

If a fulfilled promise is passed to `transitionWhile` (due to the page data being available synchronously), it goes like this:

1. Alter DOM
1. `transitionWhile` called - capture & store scroll position
1. Restore scrolling if applicable

This behaviour is broken, as the new DOM may drastically alter the scroll position. Also, the developer may choose to reset the scroll position after altering the DOM, if the page is conceptually new.

It feels like the API needs to change so it knows when to capture the scroll position.

---

An alternative design:

```js
navigation.addEventListener('navigate', (event) => {
  event.transitionWhile(async (transition) => {
    const response = await fetch(dataURL);
    const data = await response.json();
    transition.captureState();
    await updateDOM(data);
  });
});
```

Here, `transitionWhile` takes a callable. It captures scroll and focus state before calling the callable. This solves the synchronous issue.

The developer can call `captureState` to recapture state at a later time, ideally just before updating the DOM.

Scroll positions are restored when the promise returned by the callable fulfills.

domenic:
Fixes #230. Mini-explainer:

The `"navigate"` event provided by the navigation API has a method on its `NavigateEvent` instance, `event.transitionWhile(promise, options)`. This method would intercept the navigation, convert it into a same-document navigation, and then use the provided promise to signal to the browser and to other parts of the web application that we were in a transitional phase while the new navigation finishes up.

However, the signature of `transitionWhile(promise)` worked poorly in practice. The most important reason is the common coding pattern of:

```js
event.transitionWhile((async () => {
  doSyncStuff();
  await doAsyncStuff();
})());
```

i.e., the common pattern of generating a promise via an immediately-invoked async function. The problem with this pattern is that it's equivalent to

```js
doSyncStuff();
event.transitionWhile((async () => {
  await doAsyncStuff();
})());
```

This means some portion of your "navigation handling code" would run, _before_ the `navigate` event had finished firing and the URL/entry had been updated.

We know of at least two bad examples of this:

- When you synchronously modify the DOM. This throws off scroll restoration logic, as discussed in #230 (with detail at https://github.com/WICG/navigation-api/issues/230#issuecomment-1137891972). Because the sync code is modifying the DOM of the starting entry, not the destination entry.

- When you are trying to create a redirect. See https://github.com/WICG/navigation-api/issues/124#issuecomment-869804711. If you can synchronously determine you want to redirect, you need to do one behavior (canceling the current navigation, then starting a new one), whereas if you need to use async logic, then you need to do another behavior (replacing the current page). We proposed creating a first-class `navigation.transition.redirect()` helper to abstract this logic, but it's not a great sign that we'd need to create such a helper in the first place.

**Our solution is to replace `transitionWhile(promise, options)` with `intercept({ handler, ...options })`.** Essentially, we are replacing the idea of you passing a promise---which people would often generate with an async function---with the idea of you passing an async function directly. Then, the user agent can be sure to schedule the entirety of the async function _after_ the `navigate` event has finished firing, and the URL/entry updated. This avoids the above problems.

This particular API shape has other minor benefits:

- Code such as
  
  ```js
  event.transitionWhile((async () => {
    // ... many lines here ...
  })(), { focusReset: "manual" });
  ```

  hid some fairly-important information (the `focusReset: "manual"`) down at the bottom. With the new structure, we can instead write

  ```js
  event.intercept({
    focusReset: "manual",
    async handler() {
      // ... many lines here ...
    }
  });
  ```

- It is common in applications that are just starting to use the navigation API, and are not yet ready to centralize their routing logic into the `"navigate"` event handler, to simply want to intercept the navigation and convert it into a navigation-API-handled one, but leave the rest of their application as the one that reacts to this navigation. Previously this would require

  ```js
  event.transitionWhile(Promise.resolve());
  ```

  which felt like a hack. ("I want to transition, while an already-fulfilled promise is outstanding"!?) Now it is simply

  ```js
  event.intercept();
  ```

  which communicates the intent better.

- The immediately-invoked async function pattern was ugly. So ugly we were contemplating JavaScript language and Web IDL-level changes to get rid of it. See #40 and https://github.com/whatwg/webidl/issues/1020. These changes are now not necessary, at least for our API.

- This change slightly reduces (but does not fully remove) the confusion between the navigation API's notion of a navigation transition, and the in-progress shared element transitions spec's idea of a navigation transition. See https://github.com/WICG/shared-element-transitions/issues/143. The reason it does not eliminate it is that `navigation.transition` still exists. (We may settle this by renaming `navigation.transition`; see https://github.com/WICG/navigation-api/issues/230#issuecomment-1145418960.)

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/navigation-api/pull/235.html" title="Last updated on Jun 15, 2022, 4:58 PM UTC (e057298)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/navigation-api/235/388d56f...e057298.html" title="Last updated on Jun 15, 2022, 4:58 PM UTC (e057298)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,13 @@ An application or framework's centralized router can use the `navigate` event to
 
 ```js
 navigation.addEventListener("navigate", e => {
-  if (!e.canTransition || e.hashChange || e.downloadRequest !== null) {
+  if (!e.canIntercept || e.hashChange || e.downloadRequest !== null) {
     return;
   }
 
   if (routesTable.has(e.destination.url)) {
     const routeHandler = routesTable.get(e.destination.url);
-    e.transitionWhile(routeHandler());
+    e.intercept({ handler: routeHandler });
   }
 });
 ```
@@ -272,9 +272,9 @@ All of these methods return `{ committed, finished }` pairs, where both values a
 
 - It's not possible to navigate to the given entry, e.g. `navigation.traverseTo(key)` was given a non-existant `key`, or `navigation.back()` was called when there's no previous entries in the list of accessible history entries. In this case, both promises reject with an `"InvalidStateError"` `DOMException`, and `location.href` and `navigation.currentEntry` stay on their original value.
 
-- The `navigate` event responds to the navigation using `event.transitionWhile()`. In this case the `committed` promise immediately fulfills, while the `finished` promise fulfills or rejects according to the promise(s) passed to `transitionWhile()`. (However, even if the `finished` promise rejects, `location.href` and `navigation.currentEntry` will change.)
+- The `navigate` event responds to the navigation using `event.intercept()`. In this case the `committed` promise immediately fulfills, while the `finished` promise fulfills or rejects according to any promise(s) returned by handlers passed to `intercept()`. (However, even if the `finished` promise rejects, `location.href` and `navigation.currentEntry` will change.)
 
-- The navigation succeeds, and was a same-document navigation (but not intercepted using `event.transitionWhile()`). Then both promises immediately fulfill,  and `location.href` and `navigation.currentEntry` will have been set to their new value.
+- The navigation succeeds, and was a same-document navigation (but not intercepted using `event.intercept()`). Then both promises immediately fulfill,  and `location.href` and `navigation.currentEntry` will have been set to their new value.
 
 - The navigation succeeds, and it was a different-document navigation. Then the promise will never settle, because the entire document and all its promises will disappear.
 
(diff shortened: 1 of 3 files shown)


natechapin (approved):
Looks good! Just one optional nitpick.

jakearchibald:
I know I caused a lot of trouble by whinging about this late in the process, but I'm really grateful that this change happened. I'm giving a talk on this later, and this API makes it so much easier to explain (especially as I'm talking about page transitions in the same talk).

PR #248, opened 2022-09-29
Explainer for cancellable traversals

natechapin:


README.md
```diff
@@ -480,18 +480,19 @@ Similarly, cross-document navigations initiated from other windows are not somet
...
 - User-initiated traversals via the browser's back/forward buttons (either same- or cross-document)
 - Programmatic traversals via `history.back()`/`history.forward()`/`history.go()`
-- Programmatic traversals via `navigation.back()`/`navigation.forward()`/`navigation.go()`
+- Programmatic traversals via `navigation.back()`/`navigation.forward()`/`navigation.traverseTo()`
 
-We would like to make these cancelable in the future. However, we need to take care when doing so:
+Traversals may only be cancelled (and `event.cancelable` will be equal to true) if:
+- The navigate event is firing in the top window
+- The traversal is same-origin
+- The traversal was not user-initiated, or there is a consumbable user activation in the current window.
```
domenic, line 491: I think we should write the explainer toward the desired end state, of "consumable sticky activation". This will require a few sentences or a paragraph of extra explanation, but it seems better than putting something in the explainer we don't plan to ship.

README.md
```diff
@@ -480,18 +480,19 @@ Similarly, cross-document navigations initiated from other windows are not somet
 
 As for `document.open()`, it is a terrible legacy API with lots of strange side effects, which makes supporting it not worth the implementation cost. Modern sites which use the new navigation API should never be using `document.open()`.
 
-Second, the following navigations **cannot be canceled** using `event.preventDefault()`, and as such will have `event.cancelable` equal to false:
-
+Second, **traversals have special restrictions on cancelling the navigation** via `event.preventDefault()`. Traversals are:
```
domenic, line 483: Nit: I generally prefer one "l" in cancel-related conjugations, and we've codified this for the WHATWG: https://whatwg.org/style-guide .

README.md
```diff
@@ -480,18 +480,19 @@ Similarly, cross-document navigations initiated from other windows are not somet
...
+In order to enable cancellation, traversals need to fire the `navigate` event at a precise time. Most navigations fire `navigate` at the time of navigation start, but that is not a viable time for traversals, because browser architecture may require an async step to determine which frames must navigate as part of the traversal (and therefore which frames need a `navigate` event). Alternately, we could fire at `unload` time, but that is very late: by then, network requests have already been performed. `beforeunload` time splits the difference: after determining which frames will be navigated, but before any network requests or other side effects have happened. The downside of this timing is that it builds on the rickety foundation of `beforeunload` (which is widely considered to be a regrettable web platform feature). However, we believe we are avoiding the most problematic part of `beforeunload`, which is the modal user-facing dialog it creates. Simply reusing the internal browser and spec architecture for firing `navigate` at the same time as `beforeunload` is more benign.
```
domenic, line 495:
```suggestion
In order to enable cancellation, traversals need to fire the `navigate` event at a precise time. Most navigations fire `navigate` at the time of navigation start, but that is not a viable time for traversals, because browser architecture may require an async step to determine which frames must navigate as part of the traversal (and therefore which frames need a `navigate` event). Alternately, we could fire at `unload` time, but that is very late: by then, network requests have already been performed. `beforeunload` time splits the difference: after determining which frames will be navigated, but before any network requests or other side effects have happened. The downside of this timing is that it builds on the rickety foundation of `beforeunload` (which is widely considered to be a regrettable web platform feature). However, we are avoiding the most problematic part of `beforeunload`, which is the modal user-facing dialog it creates. Simply reusing the internal browser and spec architecture for firing `navigate` at the same time as `beforeunload` is more benign.
```

README.md
```diff
@@ -480,18 +480,19 @@ Similarly, cross-document navigations initiated from other windows are not somet
...
 
 Finally, the following navigations **cannot be replaced with same-document navigations** by using `event.intercept()`, and as such will have `event.canIntercept` equal to false:
```
domenic, line 498: Below here there's a paragraph starting "We'll note that these restrictions still allow canceling cross-origin non-back/forward navigations." which maybe needs some updates. Perhaps just removing the word "still" would work?

domenic (commented):
A few nits, and the biggest feedback being let's explainer/spec the consumable sticky activation.

README.md
natechapin: I think it's correct as is, but "still" isn't adding much, so i've removed it.

README.md
natechapin: Done

README.md
natechapin: Done

README.md
natechapin: Added a paragraph

README.md
```diff
@@ -480,25 +480,28 @@ Similarly, cross-document navigations initiated from other windows are not somet
...
-- Both user-initiated and programmatic traversals of this sort are hard to intercept for technical reasons, as doing so can require cross-process communication.
+By _consumable  activation_, we mean a variant of [user activation](https://html.spec.whatwg.org/multipage/interaction.html#tracking-user-activation) that we wish to add to the HTML spec. _sticky activation_ is obviously not correct for preventing trapping the user, becuase then a single errant click or tap could disable back/forward navigations entirely. However, we are also concerned about using _transient activation_: it meets our requirement that the user activation can be used once before it is consumed, but the possibility of it expiring due to its _transient activation duration_ elapsing means that web applications may suddenly and unexpectedly get an uncancelable traversal if a back or forward button is pressed and the user happens not to have interacted with the page for a modest period of time. We therefore intend to add a third mode of user activation to the HTML spec, _consumable activation_, which can be consumed like _transient activation_, but does not expire based on the _transient activation duration_.
```
domenic, line 495:
```suggestion
By _consumable activation_, we mean a variant of [user activation](https://html.spec.whatwg.org/multipage/interaction.html#tracking-user-activation) that we wish to add to the HTML spec. _Sticky activation_ is obviously not correct for preventing trapping the user, because then a single errant click or tap could disable back/forward navigations entirely. However, we are also concerned about using _transient activation_: it meets our requirement that the user activation can be used once before it is consumed, but the possibility of it expiring due to its _transient activation duration_ elapsing means that web applications may suddenly and unexpectedly get an uncancelable traversal if a back or forward button is pressed and the user happens not to have interacted with the page for a modest period of time. We therefore intend to add a third mode of user activation to the HTML spec, _consumable activation_, which can be consumed like _transient activation_, but does not expire based on the _transient activation duration_.
```

domenic (approved):
LGTM with nits (which I will fix and merge)

Issue #241, opened 2022-06-16
Handlers run at the wrong time for history traversal


domenic:
https://wicg.github.io/navigation-api/#navigate-event-traversal-patches fires the navigate event, which synchronously runs the handlers, as the very first thing during a traversal.

But this means the handlers run before the actual entry change, which is inconsistent with push/replace cases. https://whatpr.org/html/6315/history.html#apply-the-history-step step 14.10.3 should execute before we run the handlers.

PR #250, opened 2022-10-06
Cancellable traversal spec

natechapin:
***
<a href="https://pr-preview.s3.amazonaws.com/natechapin/navigation-api/pull/250.html" title="Last updated on Oct 25, 2022, 5:48 AM UTC (c59b973)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/navigation-api/250/22394af...natechapin:c59b973.html" title="Last updated on Oct 25, 2022, 5:48 AM UTC (c59b973)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -59,6 +59,7 @@ spec: html; urlPrefix: https://whatpr.org/html/6315/
     text: get all history steps; for: traversable navigable; url: history.html#getting-all-history-steps
     text: step; for: session history entry; url: history.html#she-step
     text: apply the history step; url: history.html#apply-the-history-step
+    text: check if unloading is user-canceled; url: history.html#checking-if-unloading-is-user-canceled
     text: getting the target history entry; url: history.html#getting-the-target-history-entry
     text: containing navigable; for: browsing context; url: browsers.html#bc-navigable
     text: navigable; url: history.html#navigable
@@ -1873,16 +1874,34 @@ Expand the section of the navigation/traversal response handling which deals wit
 <h3 id="navigate-event-traversal-patches">History traversal updates</h3>
 
 <div algorithm="apply the history step">
-  Modify the [=apply the history step=] algorithm as follows. Near the start of the algorithm, before checking if unloading is user-canceled (currently step 5), add the following steps:
-  1. Let |targetEntry| be the result of [=getting the target history entry=] given |traversable| and <var ignore>targetStep</var>.
-  1. If |targetEntry| does not equal |traversable|'s [=navigable/current session history entry=], then:
-    1. Let |continue| be the result of [=Firing a traversal navigate event=] at |traversable|'s [=navigable/active window=]'s [=Window/navigation API=] with <i>[=fire a traversal navigate event/destinationEntry=]</i> set to |targetEntry| and <i>[=fire a traversal navigate event/userInvolvement=]</i> set to <var ignore>userInvolvement</var>
-    1. If |continue| is false, abort these steps.
+  Modify the [=check if unloading is user-canceled=] algorithm as follows:
 
-  After setting navigable's ongoing navigation to "traversal" (currently step 8.3), add the following step:
-  1. If |navigable| does not equal |traversable|, then:
-    1. Let |continue| be the result of [=Firing a traversal navigate event=] at |navigable|'s [=navigable/active window=]'s [=Window/navigation API=] with <i>[=fire a traversal navigate event/destinationEntry=]</i> set to |targetEntry| and <i>[=fire a traversal navigate event/userInvolvement=]</i> set to <var ignore>userInvolvement</var>
-    1. [=Assert=]: |continue| is true.
+  Take an optional named argument <var ignore>targetStep</var>, default null.
+
+  Take an optional named argument <var ignore>userInvolvement</var>, (default "<code>[=user navigation involvement/none=]</code>").
+
+  Change step 4 from "Let totalTasks be the size of <var ignore>documents.</var>" to "Let totalTasks be the size of <var ignore>navigables</var>".
+
+  Wrap step 6 with "For each |navigable| of <var ignore>navigables</var>:"
+
+  Before the current step 6, add the following step:
+  1. Let |targetEntry| be the result of [=getting the target history entry=] given |navigable| and <var ignore>targetStep</var>.
+
+  Before the current step 6.1, add the following steps:
+  1. If |targetEntry|'s [=session history entry/origin=] is the [=same origin|same=] as |navigable|'s [=navigable/current session history entry=]'s [=session history entry/origin=]:
+    1. If the result of [=firing a traversal navigate event=] at |navigable|'s [=navigable/active window=]'s [=Window/navigation API=] with <i>[=fire a traversal navigate event/destinationEntry=]</i> set to |targetEntry| and <i>[=fire a traversal navigate event/userInvolvement=]</i> set to <var ignore>userInvolvement</var> is false:
+      1. Set <var ignore>unloadPromptCanceled</var> to true.
+      1. Increment <var ignore>completedTasks</var>.
+      1. Return.
+  1. If |targetEntry|'s [=session history entry/document=] is equal to |navigable|'s [=navigable/current session history entry=]'s [=session history entry/document=]:
+      1. Increment <var ignore>completedTasks</var>.
+      1. Return.
+
+  Modify the [=apply the history step=] algorithm as follows:
+  Remove step 4 (declaring <var ignore>navigablesCrossingDocuments</var>) and move step 6 (declaring <var ignore>changingNavigables</var> to step 4).
+  Change step 5 (which invokes [=check if unloading is user-canceled=]) to:
+  1. If <var ignore>checkForUserCancelation</var> is true:
+    1. If the result of [=checking if unloading is user-canceled=] given <var ignore>changingNavigables</var>, <var ignore>targetStep</var> and <var ignore>userInvolvement</var> is "refuse", then return.
 
 (Recall that we introduced the <var ignore>userInvolvement</var> parameter as part of [[#user-initiated-patches]].)
 
--- a/spec.bs
+++ b/spec.bs
@@ -1232,6 +1232,8 @@ A {{NavigateEvent}} has a <dfn for="NavigateEvent">did process scroll behavior</
 
 A {{NavigateEvent}} has a <dfn for="NavigateEvent">was intercepted</dfn>, a boolean, initially false.
 
+A {{NavigateEvent}} has a <dfn for="NavigateEvent">needs continue</dfn>, a boolean, initially false.
+
 A {{NavigateEvent}} has a <dfn for="NavigateEvent">navigation handler list</dfn>, which is a [=list=] of {{NavigationInterceptHandler}} callbacks, initially empty.
 
 <div algorithm>
@@ -1434,11 +1436,23 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
   1. Set |navigation|'s [=Navigation/focus changed during ongoing navigation=] to false.
   1. Set |navigation|'s [=Navigation/suppress normal scroll restoration during ongoing navigation=] to false.
   1. Let |dispatchResult| be the result of [=dispatching=] |event| at |navigation|.
-  1. Set |navigation|'s [=Navigation/ongoing navigate event=] to null.
   1. If |dispatchResult| is false:
     1. If |navigationType| is not "{{NavigationType/traverse}}" and |event|'s {{NavigateEvent/signal}} is not [=AbortSignal/aborted=], then [=finalize with an aborted navigation error=] given |navigation| and |ongoingNavigation|.
        <p class="note">If |navigationType| is "{{NavigationType/traverse}}", then we will [=finalize with an aborted navigation error=] in [=perform a navigation API traversal=].
     1. Return false.
+
+  1. If |destination|'s [=NavigationDestination/is same document=] is true and |navigationType| is "{{NavigationType/traverse}}":
+    1. Set |event|'s [=NavigateEvent/needs continue=] to true.
+    1. Return true;
+  1. Return the result of [=react to navigate event result=] given |navigation| and |event|.
+</div>
+
+<div algorithm>
+  The <dfn>react to `navigate` event result</dfn> is the following steps, given a {{Navigation}} |navigation|, {{NavigateEvent}} |event|:
+  1. Let |document| be |navigation|'s [=relevant global object=]'s [=associated document=].
+  1. Let |destination| be |event|'s {{NavigateEvent/destination}}.
+  1. Let |ongoingNavigation| be |navigation|'s [=Navigation/ongoing navigation=].
+  1. Let |navigationType| be |event|'s {{NavigateEvent/navigationType}}.
   1. Let |endResultIsSameDocument| be true if |event|'s [=NavigateEvent/was intercepted=] is true or |destination|'s [=NavigationDestination/is same document=] is true.
   1. [=Prepare to run script=] given |event|'s [=relevant settings object=].
       <div class="note" id="note-suppress-microtasks-during-navigation-events">
@@ -1465,6 +1479,7 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
        <p class="note">There is a subtle timing difference between how [=waiting for all=] schedules its success and failure steps when given zero promises versus &geq;1 promises. For most uses of [=waiting for all=], this does not matter. However, with this API, there are so many events and promise handlers which could fire around the same time that the difference is pretty easily observable: it can cause the event/promise handler sequence to vary. (Some of the events and promises involved include: {{Navigation/navigatesuccess}} / {{Navigation/navigateerror}}, {{Navigation/currententrychange}}, {{NavigationHistoryEntry/dispose}}, |ongoingNavigation|'s promises, and the {{NavigationTransition/finished|navigation.transition.finished}} promise.)
     1. [=Wait for all=] of |promisesList|, with the following success steps:
         1. If |event|'s {{NavigateEvent/signal}} is [=AbortSignal/aborted=], then abort these steps.
+        1. If |event| equals |navigation|'s [=Navigation/ongoing navigate event=], set |navigation|'s [=Navigation/ongoing navigate event=] to null.
         1. [=Fire an event=] named {{Navigation/navigatesuccess}} at |navigation|.
         1. If |navigation|'s [=Navigation/transition=] is not null, then [=resolve=] |navigation|'s [=Navigation/transition=]'s [=NavigationTransition/finished promise=] with undefined.
         1. Set |navigation|'s [=Navigation/transition=] to null.
@@ -1484,6 +1499,12 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
      <p class="note">Per the <a href="#note-suppress-microtasks-during-navigation-events">previous note</a>, this stops suppressing any potential promise handler microtasks, causing them to run at this point or later.</p>
   1. If |event|'s [=NavigateEvent/was intercepted=] is true and |navigationType| is "{{NavigationType/push}}", "{{NavigationType/replace}}", or "{{NavigationType/reload}}", then return false.
   1. Return true.
+</div>
+
+<div algorithm>
+  The <dfn>maybe continue the `navigate` event</dfn> is the following steps, given a {{Navigation}} |navigation|:
+  1. If |navigation|'s [=Navigation/ongoing navigate event=] is null, or |navigation|'s [=Navigation/ongoing navigate event=]'s [=NavigateEvent/needs continue=] is false, return true.
+  1. Return the result of [=react to navigate event result=] given |navigation| and |navigation|'s [=Navigation/ongoing navigate event=].
 </div>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1844,11 +1867,26 @@ Expand the section of the navigation/traversal response handling which deals wit
...
+  1. If <i>navigable</i> does not equal <i>traversable</i>, then:
+    1. Let <i>continue</i> be the result of [=Firing a traversal navigate event=] at <i>navigable</i>'s [=navigable/active window=]'s [=Window/navigation API=] with <i>[=fire a traversal navigate event/destinationEntry=]</i> set to <i>targetEntry</i> and <i>[=fire a traversal navigate event/userInvolvement=]</i> set to <var ignore>userInvolvement</var>
+    1. [=Assert=]: <i>continue</i> is true.
 
 (Recall that we introduced the <var ignore>userInvolvement</var> parameter as part of [[#user-initiated-patches]].)
+
+  Before the update document for history step application (currently step 14.10.3), add the following step:
+  1. If <i>targetEntry</i>'s [=document=] is equal to <i>displayedDocument</i>:
+    1. Let <i>updateDocument</i> be the result of [=maybe continue the navigate event=] given <i>displayedDocument</i>'s [=relevant global object=]'s [=Window/navigation API=].
+    1. If <i>updateDocument</i> is false:
```
domenic, line 1886: I think this will never be false? For the traverse case, "react to navigate event result" always returns true. So "maybe continue the navigate event" probably doesn't need a return value? And we shouldn't skip any jobs.

Basically, the return value of "react to navigate event result" is used for determining whether to stop a cross-document non-traversal nav. For the traversal case, we only allow intercepting if it's a same-document traversal anyway, so we should just let the normal same-document traversal stuff happen.

domenic (commented):
Great work! This will also fix #241.

spec.bs
natechapin: It's possible I've misunderstood, but in the traversal case, I thought we also needed to ensure that "update document for history application step" does *not* run when the traversal was intercepted. That was my goal in reading the return value here. I've update the return value of "react to navigate event result" to essentially mean "was not intercepted" to ensure we're debating what I actually meant to write. :)

spec.bs
```diff
@@ -1406,9 +1408,11 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
...
     1. Return true.
+  1. Let |browsingContext| be |navigation|'s [=relevant global object=]'s [=browsing context=].
   1. Let |document| be |navigation|'s [=relevant global object=]'s [=associated document=].
   1. If |document| <a spec="HTML">can have its URL rewritten</a> to |destination|'s [=NavigationDestination/URL=], and either |destination|'s [=NavigationDestination/is same document=] is true or |navigationType| is not "{{NavigationType/traverse}}", then initialize |event|'s {{NavigateEvent/canIntercept}} to true. Otherwise, initialize it to false.
-  1. If |navigationType| is not "{{NavigationType/traverse}}", then initialize |event|'s {{Event/cancelable}} to true. Otherwise, initialize it to false.
+  1. If |navigationType| is not "{{NavigationType/traverse}}", or if |browsingContext| is a top-level browsing context and |userInvolvement| is <code>[=user navigation involvement/browser UI=]</code> or |navigation|'s [=relevant global object=] has consumable user activation, then initialize |event|'s {{Event/cancelable}} to true. Otherwise, initialize it to false.
```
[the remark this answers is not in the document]
natechapin: Done

spec.bs
```diff
@@ -1432,11 +1436,23 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
...
   1. Let |dispatchResult| be the result of [=dispatching=] |event| at |navigation|.
-  1. Set |navigation|'s [=Navigation/ongoing navigate event=] to null.
   1. If |dispatchResult| is false:
     1. If |navigationType| is not "{{NavigationType/traverse}}" and |event|'s {{NavigateEvent/signal}} is not [=AbortSignal/aborted=], then [=finalize with an aborted navigation error=] given |navigation| and |ongoingNavigation|.
        <p class="note">If |navigationType| is "{{NavigationType/traverse}}", then we will [=finalize with an aborted navigation error=] in [=perform a navigation API traversal=].
     1. Return false.
+
+  1. If |destination|'s [=NavigationDestination/is same document=] is true and |navigationType| is "{{NavigationType/traverse}}":
+    1. Set |event|'s [=NavigateEvent/needs continue=] to true.
+    1. Return true;
```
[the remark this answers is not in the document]
natechapin: Done

spec.bs
```diff
@@ -1432,11 +1436,23 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
...
+  1. Return the result of [=react to navigate event result=] given |navigation| and |event|.
+</div>
+
+<div algorithm>
+  The <dfn>react to `navigate` event result</dfn> is the following steps, given a {{Navigation}} |navigation|, {{NavigateEvent}} |event|:
```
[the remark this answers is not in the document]
natechapin: Done

spec.bs
```diff
@@ -1484,6 +1501,12 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
   1. Return true.
 </div>
 
+<div algorithm>
+  The <dfn>maybe continue the `navigate` event</dfn> is the following steps, given a {{Navigation}} |navigation|:
```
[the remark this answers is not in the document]
natechapin: Done

spec.bs
```diff
@@ -1484,6 +1501,12 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
...
+  1. If |navigation|'s [=Navigation/ongoing navigate event=] is null, or |navigation|'s [=Navigation/ongoing navigate event=]'s [=NavigateEvent/needs continue=] is false, return true.
+  1. Return the result of [=react to navigate event result=] given |navigation| and |navigation|'s [=Navigation/ongoing navigate event=].
```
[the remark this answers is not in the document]
natechapin: Done

spec.bs
```diff
@@ -1844,11 +1867,26 @@ Expand the section of the navigation/traversal response handling which deals wit
 <h3 id="navigate-event-traversal-patches">History traversal updates</h3>
 
 <div algorithm="apply the history step">
-  Modify the [=apply the history step=] algorithm as follows. Near the end of the algorithm, inside the queued task over each global (currently step 14.10), before potentially unloading or activating the history entry, add the following step:
+  Modify the [=apply the history step=] algorithm as follows. Near the start of the algorithm, before checking if unloading is user-canceled (currently step 5), add the following steps:
+  1. Let <i>targetEntry</i> be the result of [=getting the target history entry=] given <i>traversable</i> and <i>targetStep.</i>
```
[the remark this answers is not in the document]
natechapin: Done

spec.bs
```diff
@@ -59,6 +59,7 @@ spec: html; urlPrefix: https://whatpr.org/html/6315/
     text: get all history steps; for: traversable navigable; url: history.html#getting-all-history-steps
     text: step; for: session history entry; url: history.html#she-step
     text: apply the history step; url: history.html#apply-the-history-step
```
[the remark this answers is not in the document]
natechapin: Done

spec.bs
```diff
@@ -1844,11 +1867,26 @@ Expand the section of the navigation/traversal response handling which deals wit
...
+  After setting navigable's ongoing navigation to "traversal" (currently step 8.3), add the following step:
+  1. If <i>navigable</i> does not equal <i>traversable</i>, then:
+    1. Let <i>continue</i> be the result of [=Firing a traversal navigate event=] at <i>navigable</i>'s [=navigable/active window=]'s [=Window/navigation API=] with <i>[=fire a traversal navigate event/destinationEntry=]</i> set to <i>targetEntry</i> and <i>[=fire a traversal navigate event/userInvolvement=]</i> set to <var ignore>userInvolvement</var>
+    1. [=Assert=]: <i>continue</i> is true.
 
 (Recall that we introduced the <var ignore>userInvolvement</var> parameter as part of [[#user-initiated-patches]].)
+
+  Before the update document for history step application (currently step 14.10.3), add the following step:
+  1. If <i>targetEntry</i>'s [=document=] is equal to <i>displayedDocument</i>:
```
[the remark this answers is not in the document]
natechapin: Done

spec.bs
```diff
@@ -1844,11 +1867,26 @@ Expand the section of the navigation/traversal response handling which deals wit
 <h3 id="navigate-event-traversal-patches">History traversal updates</h3>
 
 <div algorithm="apply the history step">
-  Modify the [=apply the history step=] algorithm as follows. Near the end of the algorithm, inside the queued task over each global (currently step 14.10), before potentially unloading or activating the history entry, add the following step:
+  Modify the [=apply the history step=] algorithm as follows. Near the start of the algorithm, before checking if unloading is user-canceled (currently step 5), add the following steps:
+  1. Let <i>targetEntry</i> be the result of [=getting the target history entry=] given <i>traversable</i> and <i>targetStep.</i>
+  1. If <i>targetEntry</i> does not equal <i>traversable</i>'s [=navigable/current session history entry=], then:
+    1. Let <i>continue</i> be the result of [=Firing a traversal navigate event=] at <i>traversable</i>'s [=navigable/active window=]'s [=Window/navigation API=] with <i>[=fire a traversal navigate event/destinationEntry=]</i> set to <i>targetEntry</i> and <i>[=fire a traversal navigate event/userInvolvement=]</i> set to <var ignore>userInvolvement</var>
```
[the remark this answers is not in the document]
natechapin: I've switched to this model, but I'm not sure it's entirely right. The cure draft implementation intentionally fires the navigate event for the traversable navigable first, so that the navigate event doesn't fire anywhere else if the traversal has been aborted.

spec.bs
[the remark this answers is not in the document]
domenic: Doesn't seem done:

```suggestion
    text: apply the history step; url: browsing-the-web.html#apply-the-history-step
```

spec.bs
```diff
@@ -1406,9 +1410,16 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
...
-  1. If |navigationType| is not "{{NavigationType/traverse}}", then initialize |event|'s {{Event/cancelable}} to true. Otherwise, initialize it to false.
+  1. Let |traverseCanBeCanceled| be true if |userInvolvement| is not "<code>[=user navigation involvement/browser UI=]</code>" or |navigation|'s [=relevant global object=] has [=transient activation=]; otherwise, false.
+     <p class="note">We are exploring the possibility of introducing the concept of a "consumable user activation" (i.e., a user activation that can be consumed like a transient user activation, but does not have the timeout of a transient user activation) instead of using transient activation here. This does not have a spec yet.
+  1. If all of the following are true:
+    * |navigationType| is not "{{NavigationType/traverse}}";
+    * |browsingContext| is a top-level browsing context and |traverseCanBeCanceled| is true;
```
domenic, line 1420:
```suggestion
    * |browsingContext| is a [=top-level browsing context=] and |traverseCanBeCanceled| is true
```

spec.bs
```diff
@@ -1406,9 +1410,16 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
...
   1. If |document| <a spec="HTML">can have its URL rewritten</a> to |destination|'s [=NavigationDestination/URL=], and either |destination|'s [=NavigationDestination/is same document=] is true or |navigationType| is not "{{NavigationType/traverse}}", then initialize |event|'s {{NavigateEvent/canIntercept}} to true. Otherwise, initialize it to false.
-  1. If |navigationType| is not "{{NavigationType/traverse}}", then initialize |event|'s {{Event/cancelable}} to true. Otherwise, initialize it to false.
+  1. Let |traverseCanBeCanceled| be true if |userInvolvement| is not "<code>[=user navigation involvement/browser UI=]</code>" or |navigation|'s [=relevant global object=] has [=transient activation=]; otherwise, false.
+     <p class="note">We are exploring the possibility of introducing the concept of a "consumable user activation" (i.e., a user activation that can be consumed like a transient user activation, but does not have the timeout of a transient user activation) instead of using transient activation here. This does not have a spec yet.
+  1. If all of the following are true:
+    * |navigationType| is not "{{NavigationType/traverse}}";
```
domenic, line 1419:
```suggestion
    * |navigationType| is not "{{NavigationType/traverse}}"; and
```

spec.bs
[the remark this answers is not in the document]
domenic: Hmm OK, yes, I think I gave bad advice here. So how exactly does the task queuing and thread-hopping work in the implementation? I can think of a few variants, and it's probably easier to ask you than to decipher the code...

1.
  - Browser process queues a task to the main Window to fire the navigate event + beforeunload
  - If not canceled, the browser process queues tasks to all other Windows to fire navigate event + beforeunload
1.
  - Browser process queues a task to the main Window to fire the navigate event
  - If not canceled, the browser process queues tasks to all Windows to fire navigate event (with a guard to not fire it twice on the main window) + beforeunload

plus more possibilities...?

We need to reify something similar in the spec, basically.

spec.bs
domenic: I think this is maybe an intentional departure between the implementation and spec. In the spec, same-document traversals always go down the "normal path" (i.e., "update document for history application step"), and never do the URL and history update steps. Whereas in the implementation, IIRC, you have some version of the URL and history update steps which works for traversals too, and you do that instead.

spec.bs
```diff
@@ -59,8 +59,11 @@ spec: html; urlPrefix: https://whatpr.org/html/6315/
     text: get all history steps; for: traversable navigable; url: history.html#getting-all-history-steps
     text: step; for: session history entry; url: history.html#she-step
     text: apply the history step; url: history.html#apply-the-history-step
+    text: check if unloading is user-canceled; url: history.html#checking-if-unloading-is-user-canceled
```
domenic, line 62:
```suggestion
    text: check if unloading is user-canceled; url: browsing-the-web.html#checking-if-unloading-is-user-canceled
```

spec.bs
```diff
@@ -59,8 +59,11 @@ spec: html; urlPrefix: https://whatpr.org/html/6315/
...
+    text: getting the target history entry; url: history.html#getting-the-target-history-entry
```
domenic, line 63:
```suggestion
    text: getting the target history entry; url: browsing-the-web.html#getting-the-target-history-entry
```

spec.bs
```diff
@@ -1406,9 +1410,16 @@ The <dfn attribute for="NavigationDestination">sameDocument</dfn> getter steps a
...
   1. Let |document| be |navigation|'s [=relevant global object=]'s [=associated document=].
   1. If |document| <a spec="HTML">can have its URL rewritten</a> to |destination|'s [=NavigationDestination/URL=], and either |destination|'s [=NavigationDestination/is same document=] is true or |navigationType| is not "{{NavigationType/traverse}}", then initialize |event|'s {{NavigateEvent/canIntercept}} to true. Otherwise, initialize it to false.
-  1. If |navigationType| is not "{{NavigationType/traverse}}", then initialize |event|'s {{Event/cancelable}} to true. Otherwise, initialize it to false.
+  1. Let |traverseCanBeCanceled| be true if |userInvolvement| is not "<code>[=user navigation involvement/browser UI=]</code>" or |navigation|'s [=relevant global object=] has [=transient activation=]; otherwise, false.
+     <p class="note">We are exploring the possibility of introducing the concept of a "consumable user activation" (i.e., a user activation that can be consumed like a transient user activation, but does not have the timeout of a transient user activation) instead of using transient activation here. This does not have a spec yet.
+  1. If all of the following are true:
```
natechapin: Oops, this was supposed to be an "any", not an "all".

spec.bs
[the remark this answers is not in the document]
natechapin: The implementation does the first option. The main Window fires navigate+beforeunload, and if allowed to proceed, queues tasks for all other Windows to run navigate+beforeunload as needed.

I've added a separate step for the main Window, lmk what you think of that approach.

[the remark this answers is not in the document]
domenic: OK, getting there, but...

- The patches have gotten too excessive :(. Can you instead rewrite the whole "check if unloading is user-canceled" algorithm, perhaps using something like "(Insert current steps 6.1-6.8 here)" for the parts that do stay the same? And maybe rename it to "check if unloading is canceled" at the same time, since it's not just users now...

- Instead of passing in all the navigables, then looping over them all to find the one with null parent, you can just pass in _traversable_, which is exactly the right one.

- "get all navigables that might experience a cross-document traversal" does some important work, and I'm not 100% sure we can replace it with "get all navigables whose current session history entry will change or reload". Can you walk me through what you did there?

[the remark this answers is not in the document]
natechapin: Switched to a full rewrite.

Now takes "get all navigables that might experience a cross-document traversal" and _traversable_ as parameters. Still needs all the changing navigables internally, but it now does that by running "get all navigables whose current session history entry will change or reload" internally if _traversable_ is non-null, and otherwise just copying the passed-in "get all navigables that might experience a cross-document traversal".

spec.bs
```diff
@@ -1843,12 +1875,57 @@ Expand the section of the navigation/traversal response handling which deals wit
 
 <h3 id="navigate-event-traversal-patches">History traversal updates</h3>
 
+<div algorithm="check if unloading is user-canceled">
+  Modify the [=check if unloading is user-canceled=] algorithm as follows:
+
+  To [=check if unloading is user-canceled=] given a [=list=] of [=navigables=] |navigablesCrossingDocuments|, an optional integer |targetStep| (default null), and an optional [=user navigation involvement=] |userInvolvement| (default "<code>[=user navigation involvement/none=]</code>"):
```
domenic, line 1881: Needs to take _traversable_ as an input

spec.bs
```diff
@@ -1843,12 +1875,57 @@ Expand the section of the navigation/traversal response handling which deals wit
...
+  1. Let |unloadPromptCanceled| be false.
+  1. Let |changingNavigables| be a [=list=] of [=navigables=], initially empty.
+  1. If |traversable| is not null and |targetStep| is not null, append the result of [=get all navigables whose current session history entry will change or reload=] given |traversable| and |targetStep| to |changingNavigables|.
+  1. Otherwise, append |navigablesCrossingDocuments| to |changingNavigables|.
+  1. Let |totalTasks| be the size of |changingNavigables|.
+  1. Let |completedTasks| be 0.
+  1. If |traversable| is not null and |changingNavigables| [=list/contains=] |traversable|:
+    1. Let |targetEntry| be the result of [=getting the target history entry=] given |traversable| and |targetStep|.
+    1. Let |isSameOrigin| be true if |targetEntry|'s [=session history entry/origin=] is the [=same origin|same=] as |traversable|'s [=navigable/current session history entry=]'s [=session history entry/origin=]; otherwise, false.
+    1. [=Queue a global task=] on the [=navigation and traversal task source=] given |traversable|'s [=navigable/active document=]'s [=relevant global object=] to run the following steps:
```
domenic, line 1892:
```suggestion
    1. [=Queue a global task=] on the [=navigation and traversal task source=] given |traversable|'s [=navigable/active window=] to run the following steps:
```

spec.bs
```diff
@@ -1843,12 +1875,57 @@ Expand the section of the navigation/traversal response handling which deals wit
...
+      1. If |isSameOrigin| is true:
+        1. If the result of [=firing a traversal navigate event=] at |traversable|'s [=navigable/active window=]'s [=Window/navigation API=] with <i>[=fire a traversal navigate event/destinationEntry=]</i> set to |targetEntry| and <i>[=fire a traversal navigate event/userInvolvement=]</i> set to |userInvolvement| is false:
```
domenic, line 1894: Let's make this two steps:

1. Let navigateEventResult be the result of firing...
2. If navigateEventResult is false, then set unloadPromptCanceled to true.

spec.bs
```diff
@@ -1843,12 +1875,57 @@ Expand the section of the navigation/traversal response handling which deals wit
...
+        1. Set |unloadPromptCanceled| to true.
+      1. If |unloadPromptCanceled| is false and |navigablesCrossingDocuments| [=list/contains=] |traversable|:
+        1. Run the steps to fire a beforeunload event (currently steps 6.1 thru 6.8 in the [=check if unloading is user-canceled=] steps).
+      1. Increment |completedTasks|.
+    1. Wait for |completedTasks| to be 1.
+    1. If |unloadPromptCanceled| is true, then return "<code>refuse</code>";
+  1. For each |navigable| of |changingNavigables|:
+    1. If |navigable| equals |traversable|, then continue.
```
domenic, line 1902:
```suggestion
    1. If |navigable| equals |traversable|, then [=continue=].
```

spec.bs
```diff
@@ -1843,12 +1875,57 @@ Expand the section of the navigation/traversal response handling which deals wit
...
+    1. Let |targetEntry| be the result of [=getting the target history entry=] given |traversable| and |targetStep|.
+    1. Let |isSameOrigin| be true if |targetEntry|'s [=session history entry/origin=] is the [=same origin|same=] as |navigable|'s [=navigable/current session history entry=]'s [=session history entry/origin=]; otherwise, false.
+    1. [=Queue a global task=] on the [=navigation and traversal task source=] given |navigable|'s [=navigable/active document=]'s [=relevant global object=] to run the following steps:
+      1. If |isSameOrigin| is true:
+        1. Let |navigateEventResult| be the result of [=firing a traversal navigate event=] at |navigable|'s [=navigable/active window=]'s [=Window/navigation API=] with <i>[=fire a traversal navigate event/destinationEntry=]</i> set to |targetEntry| and <i>[=fire a traversal navigate event/userInvolvement=]</i> set to |userInvolvement|:
```
domenic, line 1907:
```suggestion
        1. Let |navigateEventResult| be the result of [=firing a traversal navigate event=] at |navigable|'s [=navigable/active window=]'s [=Window/navigation API=] with <i>[=fire a traversal navigate event/destinationEntry=]</i> set to |targetEntry| and <i>[=fire a traversal navigate event/userInvolvement=]</i> set to |userInvolvement|.
```

spec.bs
```diff
@@ -1843,12 +1875,57 @@ Expand the section of the navigation/traversal response handling which deals wit
...
+        1. [=Assert=]: |navigateEventResult| is true.
+      1. If |navigablesCrossingDocuments| [=list/contains=] |navigable|:
+        1. Run the steps to fire a beforeunload event (currently steps 6.1 thru 6.8 in the [=check if unloading is user-canceled=] steps).
+      1. Increment |completedTasks|.
+  1. Wait for |completedTasks| to be |totalTasks|.
+  1. If |unloadPromptCanceled| is true, then return "<code>refuse</code>";
```
domenic, line 1913:
```suggestion
  1. If |unloadPromptCanceled| is true, then return "<code>refuse</code>".
```

spec.bs
```diff
@@ -1843,12 +1875,57 @@ Expand the section of the navigation/traversal response handling which deals wit
...
+  1. If |unloadPromptShown| is true, then return "<code>confirm</code>";
```
domenic, line 1914:
```suggestion
  1. If |unloadPromptShown| is true, then return "<code>confirm</code>".
```

domenic (commented):
OK, this looks great! Thanks for doing that. Just some editorial fixups, then we'll merge.

spec.bs
natechapin: Done

spec.bs
natechapin: Done

spec.bs
natechapin: Apparently we weren't referencing https://infra.spec.whatwg.org/ yet, so I added that. It's delightfully easy to link to new specs :)

spec.bs
natechapin: Done

spec.bs
natechapin: Done

spec.bs
natechapin: Done

spec.bs
natechapin: Done

spec.bs
domenic: So it turns out it's just `[=iteration/continue=]`, and then it gets auto-added for you with no need to modify the anchors block :). Pushed that fixup.

domenic (approved):
Nice, merging!

Next on the spec level, we need to work on consumable sticky activations. We can add that to this spec, I think, and then update the close watcher spec to replace https://wicg.github.io/close-watcher/#timestamp-of-last-activation-used-for-close-watchers with references to this spec.

spec.bs
domenic: FYI the reason the anchors block is full of things right now is for basically two reasons:

- We are using a bunch of non-exported concepts from HTML. Using non-exported concepts requires adding them to the anchors block.
- We are using the rewrite, which is not in the linking database.

In most other cases, no anchors block additions should be required. (Although sometimes ReSpec specs need to be added, because ReSpec is problematic.)

PR #257, opened 2022-12-10
Fix typo in event naming

romain-trotard:
Hello,

I just fix a little typo in this PR.
Thank you for this new API that is amazing :)

----

ps: I have created a W3C account after opening this PR


domenic:
Thanks very much, both for the fix and for the kind words!

PR #258, opened 2023-01-17
Fix/update Typescript types

atscott:
### refactor(types): NavigationInterceptOptions handler should return Promise<void>

Having the handler a `Promise<undefined>` makes it a little awkward to
implement, requiring an explicit `return undefined` at times.

`Promise<void>` allows for omitting an explicit return when using an
async function.

[Example](https://stackblitz.com/edit/t8zvpm?file=index.ts&view=editor)

### fix(types): Make info optional on NavigateEvent

I believe `info` should be optional on the `NavigateEvent` given that it
is optional in the `NavigateEventInit` and my experimentation revealed
that it is indeed `undefined` if not set in the `NavigationNavigateOptions`.

### DIFF
--- a/navigation_api.d.ts
+++ b/navigation_api.d.ts
@@ -129,7 +129,7 @@ interface NavigateEventInit extends EventInit {
 }
 
 interface NavigationInterceptOptions {
-  handler?: () => Promise<undefined>,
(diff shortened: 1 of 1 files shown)


domenic (approved):
Nice, thank you!

PR #278, opened 2025-01-24
Anchor extensions

absurdprofit:
This PR adds a proposal for extensions to the existing anchor (`<a>`) element. You can see the proposal in action in this [demo written in Lit](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy50cyIsImNvbnRlbnQiOiJpbXBvcnQgeyBMaXRFbGVtZW50LCBodG1sLCBub3RoaW5nIH0gZnJvbSAnbGl0JztcbmltcG9ydCB7IHByb3BlcnR5LCBjdXN0b21FbGVtZW50IH0gZnJvbSAnbGl0L2RlY29yYXRvcnMuanMnO1xuXG5AY3VzdG9tRWxlbWVudCgnZXh0ZW5kZWQtYW5jaG9yJylcbmNsYXNzIEV4dGVuZGVkQW5jaG9yIGV4dGVuZHMgTGl0RWxlbWVudCB7XG4gIEBwcm9wZXJ0eSh7IHR5cGU6IFN0cmluZyB9KSBocmVmOiBzdHJpbmcgfCB1bmRlZmluZWQ7XG4gIEBwcm9wZXJ0eSh7IHR5cGU6IFN0cmluZyB9KSByZWw6IHN0cmluZyB8IHVuZGVmaW5lZDtcbiAgQHByb3BlcnR5KHsgdHlwZTogU3RyaW5nIH0pIGtleTogc3RyaW5nIHwgdW5kZWZpbmVkO1xuICBAcHJvcGVydHkoeyB0eXBlOiBTdHJpbmcsIHJlZmxlY3Q6IHRydWUgfSkgbmF2aWdhdGVpbmZvOiBzdHJpbmcgfCB1bmRlZmluZWQ7XG4gIEBwcm9wZXJ0eSh7IHR5cGU6IFN0cmluZywgcmVmbGVjdDogdHJ1ZSB9KSBuYXZpZ2F0ZXN0YXRlOiBzdHJpbmcgfCB1bmRlZmluZWQ7XG4gIEBwcm9wZXJ0eSh7IHR5cGU6IEJvb2xlYW4sIHJlZmxlY3Q6IHRydWUgfSkgcmVsb2FkID0gZmFsc2U7XG4gIEBwcm9wZXJ0eSh7IHR5cGU6IEJvb2xlYW4sIHJlZmxlY3Q6IHRydWUgfSkgcmVwbGFjZSA9IGZhbHNlO1xuICBAcHJvcGVydHkoeyB0eXBlOiBCb29sZWFuLCByZWZsZWN0OiB0cnVlIH0pIHRyYXZlcnNlID0gZmFsc2U7XG4gIEBwcm9wZXJ0eSh7IHR5cGU6IFN0cmluZywgc3RhdGU6IHRydWUgfSkgY29tcHV0ZWRIcmVmOiBzdHJpbmcgfCBudWxsIHwgdW5kZWZpbmVkO1xuXG4gIHN0YXRpYyBmaW5kQ2xvc2VzdEVudHJ5SHJlZihocmVmOiBzdHJpbmcsIHJlbDogc3RyaW5nIHwgdW5kZWZpbmVkLCBlbnRyaWVzOiBOYXZpZ2F0aW9uSGlzdG9yeUVudHJ5W10sIGluZGV4OiBudW1iZXIpIHtcbiAgICBpZiAoIUFycmF5LmlzQXJyYXkoZW50cmllcykgfHwgaW5kZXggPCAwIHx8IGluZGV4ID49IGVudHJpZXMubGVuZ3RoIHx8IGVudHJpZXMubGVuZ3RoID09PSAxKSB7XG4gICAgICByZXR1cm4gdW5kZWZpbmVkO1xuICAgIH1cblxuICAgIGxldCBsZWZ0ID0gaW5kZXggLSAxO1xuICAgIGxldCByaWdodCA9IGluZGV4ICsgMTtcbiAgICBpZiAocmVsPy5pbmNsdWRlcygnbmV4dCcpKSBsZWZ0ID0gMDtcbiAgICBpZiAocmVsPy5pbmNsdWRlcygncHJldicpKSByaWdodCA9IGVudHJpZXMubGVuZ3RoO1xuXG4gICAgd2hpbGUgKGxlZnQgPj0gMCB8fCByaWdodCA8IGVudHJpZXMubGVuZ3RoKSB7XG4gICAgICBpZiAobGVmdCA-PSAwICYmIGVudHJpZXNbbGVmdF0_LnVybCA9PT0gaHJlZikgcmV0dXJuIGVudHJpZXNbbGVmdF07XG4gICAgICBpZiAocmlnaHQgPCBlbnRyaWVzLmxlbmd0aCAmJiBlbnRyaWVzW3JpZ2h0XT8udXJsID09PSBocmVmKSByZXR1cm4gZW50cmllc1tyaWdodF07XG5cbiAgICAgIGxlZnQtLTtcbiAgICAgIHJpZ2h0Kys7XG4gICAgfVxuXG4gICAgcmV0dXJuIHVuZGVmaW5lZDtcbiAgfVxuXG4gIHN0YXRpYyBmaW5kQ2xvc2VzdEVudHJ5KHJlbDogc3RyaW5nIHwgdW5kZWZpbmVkLCBlbnRyaWVzOiBOYXZpZ2F0aW9uSGlzdG9yeUVudHJ5W10sIGluZGV4OiBudW1iZXIpIHtcbiAgICBpZiAoIUFycmF5LmlzQXJyYXkoZW50cmllcykgfHwgaW5kZXggPCAwIHx8IGluZGV4ID49IGVudHJpZXMubGVuZ3RoIHx8IGVudHJpZXMubGVuZ3RoID09PSAxKSB7XG4gICAgICByZXR1cm4gdW5kZWZpbmVkO1xuICAgIH1cblxuICAgIGxldCBsZWZ0ID0gLTE7XG4gICAgbGV0IHJpZ2h0ID0gZW50cmllcy5sZW5ndGg7XG4gICAgaWYgKHJlbD8uaW5jbHVkZXMoJ3ByZXYnKSkgbGVmdCA9IGluZGV4IC0gMTtcbiAgICBpZiAocmVsPy5pbmNsdWRlcygnbmV4dCcpKSByaWdodCA9IGluZGV4ICsgMTtcblxuICAgIHdoaWxlIChsZWZ0ID49IDAgfHwgcmlnaHQgPCBlbnRyaWVzLmxlbmd0aCkge1xuICAgICAgaWYgKGxlZnQgPj0gMCkgcmV0dXJuIGVudHJpZXNbbGVmdF07XG4gICAgICBpZiAocmlnaHQgPCBlbnRyaWVzLmxlbmd0aCkgcmV0dXJuIGVudHJpZXNbcmlnaHRdO1xuXG4gICAgICBsZWZ0LS07XG4gICAgICByaWdodCsrO1xuICAgIH1cblxuICAgIHJldHVybiB1bmRlZmluZWQ7XG4gIH1cblxuICBjb25uZWN0ZWRDYWxsYmFjaygpIHtcbiAgICBzdXBlci5jb25uZWN0ZWRDYWxsYmFjaygpO1xuICAgIHdpbmRvdy5uYXZpZ2F0aW9uLmFkZEV2ZW50TGlzdGVuZXIoJ25hdmlnYXRlc3VjY2VzcycsIHRoaXMuX29uTmF2aWdhdGUpO1xuICB9XG4gIGRpc2Nvbm5lY3RlZENhbGxiYWNrKCkge1xuICAgIHdpbmRvdy5uYXZpZ2F0aW9uLnJlbW92ZUV2ZW50TGlzdGVuZXIoJ25hdmlnYXRlc3VjY2VzcycsIHRoaXMuX29uTmF2aWdhdGUpO1xuICAgIHN1cGVyLmRpc2Nvbm5lY3RlZENhbGxiYWNrKCk7XG4gIH1cblxuICBfb25OYXZpZ2F0ZSA9ICgpID0-IHtcbiAgICBpZiAoIXRoaXMuaHJlZiAmJiB0aGlzLnRyYXZlcnNlKSB7XG4gICAgICBsZXQgZW50cnk7XG4gICAgICBpZiAodGhpcy5rZXkpXG4gICAgICAgIGVudHJ5ID0gd2luZG93Lm5hdmlnYXRpb24uZW50cmllcygpLmZpbmQoZW50cnkgPT4gZW50cnkua2V5ID09PSB0aGlzLmtleSk7XG4gICAgICBlbHNlIGlmICh0aGlzLnJlbClcbiAgICAgICAgZW50cnkgPSBFeHRlbmRlZEFuY2hvci5maW5kQ2xvc2VzdEVudHJ5KHRoaXMucmVsLCB3aW5kb3cubmF2aWdhdGlvbi5lbnRyaWVzKCksIHdpbmRvdy5uYXZpZ2F0aW9uLmN1cnJlbnRFbnRyeT8uaW5kZXggPz8gMCk7XG5cbiAgICAgIHRoaXMuY29tcHV0ZWRIcmVmID0gZW50cnk_LnVybDtcbiAgICB9XG4gIH1cblxuICBoYW5kbGVDbGljayhldmVudDogUG9pbnRlckV2ZW50KSB7XG4gICAgZXZlbnQucHJldmVudERlZmF1bHQoKTtcbiAgICBjb25zdCBhID0gdGhpcy5yZW5kZXJSb290LmZpcnN0RWxlbWVudENoaWxkIGFzIEhUTUxBbmNob3JFbGVtZW50O1xuXG4gICAgY29uc3QgbmF2aWdhdGlvbiA9IHdpbmRvdy5uYXZpZ2F0aW9uO1xuICAgIGlmICghbmF2aWdhdGlvbikge1xuICAgICAgY29uc29sZS53YXJuKCdOYXZpZ2F0aW9uIEFQSSBub3Qgc3VwcG9ydGVkIGluIHRoaXMgYnJvd3Nlci4nKTtcbiAgICAgIHJldHVybjtcbiAgICB9XG5cbiAgICBjb25zdCB7IHJlbCwga2V5LCByZWxvYWQsIHJlcGxhY2UsIHRyYXZlcnNlLCBuYXZpZ2F0ZWluZm86IGluZm8sIG5hdmlnYXRlc3RhdGU6IHN0YXRlIH0gPSB0aGlzO1xuICAgIGNvbnN0IGhyZWYgPSBhLmhyZWY7XG5cbiAgICBpZiAodHJhdmVyc2UpIHtcbiAgICAgIGNvbnN0IGVudHJpZXMgPSBuYXZpZ2F0aW9uLmVudHJpZXMoKTtcbiAgICAgIGNvbnN0IGVudHJ5ID1cbiAgICAgICAgRXh0ZW5kZWRBbmNob3IuZmluZENsb3Nlc3RFbnRyeUhyZWYoaHJlZiwgcmVsLCBlbnRyaWVzLCBuYXZpZ2F0aW9uLmN1cnJlbnRFbnRyeT8uaW5kZXggPz8gMCkgPz9cbiAgICAgICAgRXh0ZW5kZWRBbmNob3IuZmluZENsb3Nlc3RFbnRyeShyZWwsIGVudHJpZXMsIG5hdmlnYXRpb24uY3VycmVudEVudHJ5Py5pbmRleCA_PyAwKTtcblxuICAgICAgaWYgKGtleSB8fCBlbnRyeSkge1xuICAgICAgICBuYXZpZ2F0aW9uLnRyYXZlcnNlVG8oa2V5IHx8IGVudHJ5IS5rZXksIHsgaW5mbyB9KTtcbiAgICAgICAgcmV0dXJuO1xuICAgICAgfVxuICAgICAgY29uc29sZS5sb2coeyBrZXksIGVudHJ5IH0pO1xuICAgIH1cbiAgICBjb25zb2xlLmxvZyh7IGhyZWYsIHJlbCwgc3RhdGUsIGluZm8sIHRyYXZlcnNlLCByZXBsYWNlLCByZWxvYWQgfSk7XG5cbiAgICBpZiAocmVwbGFjZSkge1xuICAgICAgaHJlZiAmJiBuYXZpZ2F0aW9uLm5hdmlnYXRlKGhyZWYsIHsgaW5mbywgc3RhdGUsIGhpc3Rvcnk6ICdyZXBsYWNlJyB9KTtcbiAgICB9IGVsc2UgaWYgKHJlbG9hZCkge1xuICAgICAgbmF2aWdhdGlvbi5yZWxvYWQoeyBpbmZvLCBzdGF0ZSB9KTtcbiAgICB9IGVsc2Uge1xuICAgICAgaHJlZiAmJiBuYXZpZ2F0aW9uLm5hdmlnYXRlKGhyZWYsIHsgaW5mbywgc3RhdGUgfSk7XG4gICAgfVxuICB9XG5cbiAgcmVuZGVyKCkge1xuICAgIGlmICh0aGlzLnJlbG9hZClcbiAgICAgIHRoaXMuaHJlZiA9ICcuJztcbiAgICByZXR1cm4gaHRtbGBcbiAgICAgIDxhXG4gICAgICAgIGhyZWY9XCIke3RoaXMuaHJlZiA_PyB0aGlzLmNvbXB1dGVkSHJlZiA_PyBub3RoaW5nfVwiXG4gICAgICAgIHJlbD1cIiR7dGhpcy5yZWwgPz8gbm90aGluZ31cIlxuICAgICAgICBrZXk9XCIke3RoaXMua2V5ID8_IG5vdGhpbmd9XCJcbiAgICAgICAgQGNsaWNrPVwiJHt0aGlzLmhhbmRsZUNsaWNrfVwiXG4gICAgICA-XG4gICAgICAgIDxzbG90Pjwvc2xvdD5cbiAgICAgIDwvYT5cbiAgICBgO1xuICB9XG59XG4ifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhRE9DVFlQRSBodG1sPlxuPGhlYWQ-XG4gIDxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vc2ltcGxlLWdyZWV0aW5nLmpzXCI-PC9zY3JpcHQ-XG48L2hlYWQ-XG48Ym9keT5cbiAgPGRpdiBpZD1cIm1haW5cIj5cbiAgICAgIDxoMT5BbmNob3IgRWxlbWVudCBFeHRlbnNpb248L2gxPlxuICAgICAgPHVsPlxuICAgICAgICA8bGk-XG4gICAgICAgICAgPGV4dGVuZGVkLWFuY2hvciBocmVmPVwiL2hlbGxvXCI-UHVzaDwvZXh0ZW5kZWQtYW5jaG9yPlxuICAgICAgICA8L2xpPlxuICAgICAgICA8bGk-XG4gICAgICAgICAgPGV4dGVuZGVkLWFuY2hvciByZWxvYWQ-UmVsb2FkPC9leHRlbmRlZC1hbmNob3I-XG4gICAgICAgIDwvbGk-XG4gICAgICAgIDxsaT5cbiAgICAgICAgICA8ZXh0ZW5kZWQtYW5jaG9yIGhyZWY9XCIvaGVsbG8td29ybGRcIiByZXBsYWNlPlJlcGxhY2U8L2V4dGVuZGVkLWFuY2hvcj5cbiAgICAgICAgPC9saT5cbiAgICAgICAgPGxpPlxuICAgICAgICAgIDxleHRlbmRlZC1hbmNob3IgcmVsPVwicHJldlwiIHRyYXZlcnNlPlRyYXZlcnNlIGJhY2s8L2V4dGVuZGVkLWFuY2hvcj5cbiAgICAgICAgPC9saT5cbiAgICAgIDwvdWw-XG5cbiAgICAgIDxuYXYgY2xhc3M9XCJ0YWJzXCI-XG4gICAgICAgIDxleHRlbmRlZC1hbmNob3IgY2xhc3M9XCJ0YWJcIiBocmVmPVwidGFiLTFcIiB0cmF2ZXJzZT5UYWIgMTwvZXh0ZW5kZWQtYW5jaG9yPlxuICAgICAgICA8ZXh0ZW5kZWQtYW5jaG9yIGNsYXNzPVwidGFiXCIgaHJlZj1cInRhYi0yXCIgdHJhdmVyc2U-VGFiIDI8L2V4dGVuZGVkLWFuY2hvcj5cbiAgICAgICAgPGV4dGVuZGVkLWFuY2hvciBjbGFzcz1cInRhYlwiIGhyZWY9XCJ0YWItM1wiIHRyYXZlcnNlPlRhYiAzPC9leHRlbmRlZC1hbmNob3I-XG4gICAgICA8L25hdj5cbiAgICA8L2Rpdj5cblxuICAgIDxzY3JpcHQ-XG4gICAgICBuYXZpZ2F0aW9uLmFkZEV2ZW50TGlzdGVuZXIoJ25hdmlnYXRlJywgKGUpID0-IHtcbiAgICAgICAgZS5pbnRlcmNlcHQoeyBoYW5kbGVyOiAoKSA9PiBjb25zb2xlLmxvZyhlKSB9KTtcbiAgICAgIH0pO1xuICAgIDwvc2NyaXB0PlxuPC9ib2R5PlxuIn0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjMuMC4wXCIsXG4gICAgXCJAbGl0L3JlYWN0aXZlLWVsZW1lbnRcIjogXCJeMi4wLjBcIixcbiAgICBcImxpdC1lbGVtZW50XCI6IFwiXjQuMC4wXCIsXG4gICAgXCJsaXQtaHRtbFwiOiBcIl4zLjAuMFwiXG4gIH1cbn0iLCJoaWRkZW4iOnRydWV9XQ).

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ backButtonEl.addEventListener("click", () => {
   - [Notifications on entry disposal](#notifications-on-entry-disposal)
   - [Current entry change monitoring](#current-entry-change-monitoring)
   - [Complete event sequence](#complete-event-sequence)
+  - [Extensions to the Anchor Element](#extensions-to-the-anchor-element)
 - [Guide for migrating from the existing history API](#guide-for-migrating-from-the-existing-history-api)
   - [Performing navigations](#performing-navigations)
   - [Warning: back/forward are not always opposites](#warning-backforward-are-not-always-opposites)
@@ -1032,6 +1033,19 @@ The commit steps are:
 1. `currententrychange` is fired on `navigation`.
 1. Any now-unreachable `NavigationHistoryEntry` instances fire `dispose`.
 1. The URL bar updates.
+
+### Extensions to the Anchor element
+Currently an `<a>` only supports push navigations, which leaves a lot to be desired. This proposal aims to extend its existing behaviour with a few new and existing attributes.
+Namely:
+- `replace="", reload=""` and `traverse=""` as boolean attributes.
+- `navigateinfo=""` and `navigatestate=""` as string attributes.
+- `href="..."` and `rel="next/prev"` with the addition of the string attribute `key="..."` as traversal hints.
+
+The default navigation type of a `<a>` is always a push navigation. The new boolean attributes `replace="", reload=""` and `traverse=""` serve as the default navigation type modifier.
+When using the `traverse=""` navigation type modifier the `rel="next/prev"`, `key="..."` and `href="..."` attributes, become optional hints.
+
+#### Default navigation type
+The `replace=""` attribute in addition to the `href="..."` attribute will replace the current entry with a new history entry with the resolved href as the URL.
 
 ## Guide for migrating from the existing history API
 
--- a/README.md
+++ b/README.md
@@ -1118,8 +1118,8 @@ Tabbed navigation could be implemented without JavaScript. By using `traverse=""
 ```html
 <nav class="tabs">
   <a class="tab" href="tab-1" traverse>Tab 1</a>
-  <a class="tab" href="tab-1" traverse>Tab 2</a>
-  <a class="tab" href="tab-1" traverse>Tab 3</a>
+  <a class="tab" href="tab-2" traverse>Tab 2</a>
+  <a class="tab" href="tab-3" traverse>Tab 3</a>
 </nav>
 ```
 
--- a/README.md
+++ b/README.md
@@ -84,7 +84,10 @@ backButtonEl.addEventListener("click", () => {
   - [Notifications on entry disposal](#notifications-on-entry-disposal)
   - [Current entry change monitoring](#current-entry-change-monitoring)
   - [Complete event sequence](#complete-event-sequence)
-  - [Extensions to the Anchor Element](#extensions-to-the-anchor-element)
+  - [Extensions to the Anchor element](#extensions-to-the-anchor-element)
+    - [Default navigation type](#default-navigation-type)
+    - [The navigateinfo="..." and navigatestate="..." attributes](#the-navigateinfo-and-navigatestate-attributes)
+    - [Additional example](#additional-example)
 - [Guide for migrating from the existing history API](#guide-for-migrating-from-the-existing-history-api)
   - [Performing navigations](#performing-navigations)
   - [Warning: back/forward are not always opposites](#warning-backforward-are-not-always-opposites)
@@ -1035,17 +1038,90 @@ The commit steps are:
 1. The URL bar updates.
 
 ### Extensions to the Anchor element
-Currently an `<a>` only supports push navigations, which leaves a lot to be desired. This proposal aims to extend its existing behaviour with a few new and existing attributes.
+Currently an `<a>` only supports push navigations. This proposal aims to extend its existing behaviour with a few new and existing attributes.
 Namely:
-- `replace="", reload=""` and `traverse=""` as boolean attributes.
+- `reload=""`, `replace=""` and `traverse=""` as boolean attributes.
 - `navigateinfo=""` and `navigatestate=""` as string attributes.
 - `href="..."` and `rel="next/prev"` with the addition of the string attribute `key="..."` as traversal hints.
 
-The default navigation type of a `<a>` is always a push navigation. The new boolean attributes `replace="", reload=""` and `traverse=""` serve as the default navigation type modifier.
-When using the `traverse=""` navigation type modifier the `rel="next/prev"`, `key="..."` and `href="..."` attributes, become optional hints.
-
 #### Default navigation type
-The `replace=""` attribute in addition to the `href="..."` attribute will replace the current entry with a new history entry with the resolved href as the URL.
+The default navigation type of a `<a>` is always a push navigation. The new boolean attributes `replace=""`, `reload=""` and `traverse=""` serve as the default navigation type modifier.
+
+The default navigation type will still be a push navigation.
+```html
+<a href="posts">Posts</a>
+```
+
+The `reload=""` attribute will reload the current entry. Note: the `href=""` attribute is ignored.
+```html
+<a reload>Refresh results</a>
+```
+
+The `replace=""` attribute in addition to the `href="..."` attribute will replace the current entry with a new history entry.
+```html
+<a href="/login" replace>Logout</a>
+```
+The `traverse=""` attibute allows `<a>`s to link to existing entries. When using the `traverse=""` navigation type modifier the `rel="next/prev"`, `key="..."` and `href="..."` attributes, become optional hints.
+The `rel="..."` is an existing attribute that hints at the relationship of the linked URL to the current document. When used with `traverse=""` it serves as a directional hint.
+```html
+<a rel="prev" traverse>Go back</a>
+<a rel="next" traverse>Go forward</a>
+```
+The `href="..."` is an existing attribute that specifies the URL the `<a>` points to. When used with the `traverse=""` attribute the anchor will point to the closest entry with a matching URL. The `rel="next/prev"` attribute can be used to specify the search direction.
+```html
+<a href=".." traverse>Tab 1</a> <!-- Given the current URL pathname is /parent/child this anchor navigates to an entry with /parent as the pathname. -->
+<a href="/" rel="prev" traverse>Go home</a>
+<a href="/posts" rel="next" traverse>Posts</a>
+```
+The `key="..."` is a new attribute that allows you to specify a history entry key the `<a>` points to. This attribute only works with the `traverse=""` attribute and is ignored otherwise. If no matching entry is found the behaviour is the same as `<a>Dead link</a>`.
+```html
+<a key="7c6a6481-c4bc-40f5-9617-0287441a3418" traverse>Go to 7c6a6481-c4bc-40f5-9617-0287441a3418</a>
+```
+The `key="..."` attribute takes precedence over `href=""` meaning if an entry with a matching key isn't found, the anchor will fallback to traversal with the `href="..."` hint.
+```html
+<a href="/" key="7c6a6481-c4bc-40f5-9617-0287441a3418" traverse>Go home</a>
+```
+
+The `traverse=""` boolean attribute will always take precedence over push, replace and reload navigation types. Meaning if no matching entry is found the `<a>` will fallback to the next navigation type.
+```html
+<a href="/" rel="prev" traverse replace>Go home</a> <!-- / Will replace the current entry if no matching entry is found -->
+```
+
+#### The `navigateinfo="..."` and `navigatestate="..."` attributes
+Finally two string attributes have been added for passing state and info via an `<a>`.
+```html
+<a id="home" href="/" navigateinfo="push" navigatestate="newuser">Home</a>
+```
+The new attributes will have corresponding DOM properties for passing non-string data types to entries.
+```js
+const a = document.getElementById("home");
+a.navigatestate = {
+  user: {
+    name: "John Doe",
+  },
+};
+a.navigateinfo = {
+  type: "push",
+};
+```
+
+Note: `navigatestate=""` is ignored for traverse navigations.
+```html
+<a href="/posts" navigateinfo="push" navigatestate="newstate">Home</a>
+<a navigateinfo="reload" navigatestate="newstate" reload>Refresh</a>
+<a href="/login" navigateinfo="replace" navigatestate="newstate" replace>Logout</a>
+<a rel="prev" navigateinfo="traverse" navigatestate="newstate" traverse>Back</a> <!-- navigatestate="newstate" is ignored -->
+```
+
+#### Additional Example
+Tabbed navigation could be implemented without JavaScript. By using `traverse=""` anchors will point to existing entries if they exist but fallback to a push navigation if a matching entry doesn't already exist.
+```html
+<nav class="tabs">
+  <a class="tab" href="tab-1" traverse>Tab 1</a>
+  <a class="tab" href="tab-1" traverse>Tab 2</a>
+  <a class="tab" href="tab-1" traverse>Tab 3</a>
+</nav>
+```
 
 ## Guide for migrating from the existing history API
 


README.md
```diff
@@ -1033,6 +1037,92 @@ The commit steps are:
...
+The `reload=""` attribute will reload the current entry. Note: the `href=""` attribute is ignored.
+```html
+<a reload>Refresh results</a>
+```
+
+The `replace=""` attribute in addition to the `href="..."` attribute will replace the current entry with a new history entry.
+```html
+<a href="/login" replace>Logout</a>
+```
+The `traverse=""` attibute allows `<a>`s to link to existing entries. When using the `traverse=""` navigation type modifier the `rel="next/prev"`, `key="..."` and `href="..."` attributes, become optional hints.
```
domenic, line 1064: Similar questions for the href-less versions here.

Additionally, to get `key=""`s you need to use JavaScript. So I can't really understand when you would use this.

I can see a use case for `<a href="/" rel="prev" traverse>`, however.

README.md
```diff
@@ -1033,6 +1037,92 @@ The commit steps are:
...
+The `rel="..."` is an existing attribute that hints at the relationship of the linked URL to the current document. When used with `traverse=""` it serves as a directional hint.
+```html
+<a rel="prev" traverse>Go back</a>
+<a rel="next" traverse>Go forward</a>
+```
+The `href="..."` is an existing attribute that specifies the URL the `<a>` points to. When used with the `traverse=""` attribute the anchor will point to the closest entry with a matching URL. The `rel="next/prev"` attribute can be used to specify the search direction.
+```html
+<a href=".." traverse>Tab 1</a> <!-- Given the current URL pathname is /parent/child this anchor navigates to an entry with /parent as the pathname. -->
+<a href="/" rel="prev" traverse>Go home</a>
+<a href="/posts" rel="next" traverse>Posts</a>
```
domenic, line 1074: What happens if there are no links in history with these hrefs?

README.md
```diff
@@ -1033,6 +1037,92 @@ The commit steps are:
...
+```js
+const a = document.getElementById("home");
+a.navigatestate = {
+  user: {
+    name: "John Doe",
+  },
+};
+a.navigateinfo = {
+  type: "push",
+};
```
domenic, line 1105: I think this gets pretty messy.

The issues mostly come about with the interaction between the JS properties and the attributes.

- What does the attribute return, when the JS property is set to an object?
- What does the JS property return, when the attribute is set to a string?
- How does `el.setAttribute("navigatestate", el.getAttribute("navigatestate"))` behave? Does it change what state is sent? Does it change the value of the JS property?
- How does `el.navigatestate = el.navigatestate` behave? Does it change what state is sent? Does it change the value of the attribute?

I would suggest leaving these as strings only and saying that if you want more complex objects, you need to use JavaScript.

README.md
```diff
@@ -1033,6 +1037,92 @@ The commit steps are:
...
+Note: `navigatestate=""` is ignored for traverse navigations.
+```html
+<a href="/posts" navigateinfo="push" navigatestate="newstate">Home</a>
+<a navigateinfo="reload" navigatestate="newstate" reload>Refresh</a>
+<a href="/login" navigateinfo="replace" navigatestate="newstate" replace>Logout</a>
+<a rel="prev" navigateinfo="traverse" navigatestate="newstate" traverse>Back</a> <!-- navigatestate="newstate" is ignored -->
+```
+
+#### Additional Example
+Tabbed navigation could be implemented without JavaScript. By using `traverse=""` anchors will point to existing entries if they exist but fallback to a push navigation if a matching entry doesn't already exist.
```
domenic, line 1117: Is this actually a user experience that people want? Are there any sites today that are implementing this kind of user experience using JavaScript?

domenic (commented):
Thanks for showing us what this would look like!

I think what is most missing here is some evidence that this kind of functionality is something developers want. Especially for the more complex pieces. Is anyone creating experiences like these today using JavaScript? Can you find live examples, ideally on popular sites? Or even just StackOverflow questions of people asking how to do this sort of thing?

README.md
absurdprofit: - The getAttribute() would return the attribute as seen in HTML.
- The JS property at first would be the same as the HTML attribute. In other words the attribute would serve as a "default value".
- No, the attribute wouldn't change the value of the JS property. The JS property takes precedent in the state that is sent.
- No, the JS property wouldn't change the value of the attribute.

TLDR; the idea here is the navigatestate and navigateinfo properties/attributes would mirror the behaviour of the value attribute/property from the `<input>` element. The attribute has no effect on the property beyond its use as a "default value".

This wasn't very clear in the demo and I apologise.

README.md
absurdprofit: If there are no links in the history then it should fallback to the base navigate behaviour. The default base navigate behaviour for `<a>`s as they exist now is "push", meaning if a link doesn't exist it simply becomes a push navigation. The proposal also adds the ability to modify this "base navigate behaviour" so for example a `replace=""` boolean attribute would make the fallback instead be a replace navigation.

README.md
absurdprofit: The `key="..."` attribute is merely a convenience instead of a Super Declarative™ API. This exists as a method for when you want to be super precise about which entry you navigate to unlike the `href="..."` pattern matching feature. Additionally having an `<a>` tag that navigates to a specific existing entry without clickjacking is essentially impossible without something like the [NavigateEvent.sourceElement](https://github.com/WICG/navigation-api/pull/264) feature.

A solid use case for something like this that I thought of is implementing breadcrumbs, most sites currently implement such a feature using the basic push navigation behaviour of the `<a>` which can lead to a messy history stack as a "previous" page now exists at the head of the history stack.

README.md
```diff
@@ -1033,6 +1037,92 @@ The commit steps are:
...
+
+The default navigation type will still be a push navigation.
+```html
+<a href="posts">Posts</a>
+```
+
+The `reload=""` attribute will reload the current entry. Note: the `href=""` attribute is ignored.
+```html
+<a reload>Refresh results</a>
+```
```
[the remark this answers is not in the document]
absurdprofit: Yea I'm not very sure either, I added it more for completeness sake. I did think it would be nice to have a "open in next tab" functionality for a refresh UI element which a `<button>` would lack, which is why I still added it. If there isn't much interest in something like this I would drop it.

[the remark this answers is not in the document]
absurdprofit: I do have a question though, what kind of navigation is a `<a href='.'>` i.e. what should NavigateEvent.navigationType yield? I tried this in chrome a few times, sometimes it would be a push navigation and sometimes it would be a replace navigation. I ask because my initial intuition was this would have been a "reload" navigation but since that wasn't the case I felt motivated to add a keyword for it.

README.md
absurdprofit: The main tabs in Gmail are implemented using JavaScript, however like many implementations online these navigations aren't added to the history stack and it's not trivial to implement synchronisation with the history stack especially with the old history API.
image

This is also the case for Azure which is an SPA that uses hash based navigation but they still don't persist these navigations to the stack.
Screenshot 2025-01-30 212946

I couldn't find any example of tab navigation by history stack traversal via JS but from my experience a lot of the anxiety around doing things like that on the web comes from the fact that it's hard. Also the default navigation behaviour of an `<a>` being push further encourages developers to use JavaScript based solutions that aren't synched to the browser history stack to avoid duplicate history entries polluting the stack (among other issues).

One problem with omitting tab navigation from the browser history is that users lose their position upon reloading the page, forcing them to manually restore their previous state. Additionally, tabs are not directly linkable, which can be frustrating if you wish to share or revisit a specific tab. To address this in the past, I have implemented a solution that synchronizes the tab state with the query parameters in the URL. This approach ensures that users can reload the page without losing their current tab selection while also making tabs linkable. At the same time, it avoids polluting the history stack with duplicate entries by updating the URL without pushing new history entries unnecessarily. By reading the initial tab state from `window.location.search`, I could restore the current tab when the page loads. [CodeSandbox implements file tabs like this](https://codesandbox.io/p/sandbox/react-complete-guide-xgxpq).

absurdprofit:
> Can you find live examples, ideally on popular sites? Or even just StackOverflow questions of people asking how to do this sort of thing?

Here's a SO post I came across once about [using only an `<a>` to implement a "back link"](https://stackoverflow.com/questions/8814472/how-to-make-an-html-back-link).

absurdprofit:
> I think what is most missing here is some evidence that this kind of functionality is something developers want.
By any chance are there any channels you recommend I use to solicit this kind of feedback?

PR #281, opened 2025-03-11
Update README.md to use precommit handlers instead of `after-transition`

noamr:


noamr:
@domenic PTAL?

README.md
```diff
@@ -691,28 +691,29 @@ Some more details on how the navigation API handles scrolling with `"traverse"`
...
+The default behavior of immediately "committing" (i.e., updating `location.href` and `navigation.currentEntry`) works well for most situations, but some developers may find they do not want to immediately update the URL, and may want to retain the option of aborting the navigation without needing to rollback a URL update or cancel-and-restart. This behavior can be customized passing a `precommitHandler` callback alongside or instead of the `handler` callback:
 
-- `e.intercept({ handler, commit: "immediate" })`: the default behavior, immediately commit the navigation and update `location.href` and `navigation.currentEntry`.
-- `e.intercept({ handler, commit: "after-transition" })`: start the navigation (e.g., show a loading spinner if the UI has one), but do not immediately commit.
+- `e.intercept({ handler })`: the default behavior, immediately commit the navigation and update `location.href` and `navigation.currentEntry`.
+- `e.intercept({ precommitHandler })`: start the navigation (e.g., show a loading spinner if the UI has one), but do not immediately commit.
```
domenic, line 699: We should probably include a bullet for what happens if you include both?

And maybe an example that uses both? (Or just modify the redirect example to separate the two parts.)

README.md
```diff
@@ -691,28 +691,29 @@ Some more details on how the navigation API handles scrolling with `"traverse"`
...
 
-When deferred commit is used, the navigation will commit (and a `committed` promise will resolve if present) when `e.commit()` is called. If any handler(s) passed to `intercept()` fulfill, and `e.commit()` has not yet been called, we will fallback to committing just before any `finish` promise resolves and `navigatesuccess` is fired.
+When precommit handlers are used, the navigation will commit (and a `committed` promise will resolve if present) once all those handlers are fulfilled. Unlike the ordinary `handler`, the `precommitHandler` callbacks are called in sequence - the next `precommit` handler is invoked only when the previous one is fulfilled. That is due to the fact that a precommit handler can asynchronously abort the navigation altogether or redirect the URL, and the next precommit handler should respond to the new state.
```
domenic, line 701:
```suggestion
When precommit handlers are used, the navigation will commit (and a `committed` promise will resolve if present) once all those handlers are fulfilled. Unlike the ordinary `handler`, the `precommitHandler` callbacks are called in sequence—the next `precommit` handler is invoked only when the previous one is fulfilled. That is due to the fact that a precommit handler can asynchronously abort the navigation altogether or redirect the URL, and the next precommit handler should respond to the new state.
```

README.md
```diff
@@ -691,28 +691,29 @@ Some more details on how the navigation API handles scrolling with `"traverse"`
...
+If a `precommitHandler` passed to `intercept()` rejects, then the navigation will be treated as canceled (both `committed` and `finished` promises will reject, and no URL update will occur).
 
-Because deferred commit can be used to cancel the navigation before the URL updates, it is only available when `e.cancelable` is true. See [above](#restrictions-on-firing-canceling-and-responding) for details on when `e.cancelable` is set to false, and thus deferred commit is not available.
+Because precommit handlers can be used to cancel the navigation before the URL updates, they are only available when `e.cancelable` is true. See [above](#restrictions-on-firing-canceling-and-responding) for details on when `e.cancelable` is set to false, and thus precommit handlers are not available.
```
domenic, line 705: We should state what happens if you try to use them. Presumably throw an exception, I guess "SecurityError" DOMException? (We did not state what would happen for commit(), but, I guess it was more obvious that since it's a method it would throw.)

domenic (commented):
LGTM with minor suggestions. This is nice!

README.md
noamr: Done

README.md
noamr: Done

noamr:
> LGTM with minor suggestions. This is nice!

Thanks! Can you add me as a contributor to this repo or approve the PR officially please :)?

